PR #715, opened 2020-09-04
test: POC of APM CLI

kuisathaverat:
## What does this PR do?

* It adds a Python script to generate transactions/span from the command line.
* It adds an example pipeline to use the script (rudimentary is a POC)

## Why is it important?

It allows us to show how can we send APM data about our CI/CD and use the same tools we use to develop to track issues in the CI.

## How to test it

* Clone the Library rep, then start a Jenkins instance and a worker
```
git clone git@github.com:elastic/apm-pipeline-library.git
cd apm-pipeline-library/local
make start
make start-local-worker
```
* From the Jenkins UI, Create a pipeline job and copy the content of `resources/testAPMCli.groovy`
* Execute the job
image

image

The job is executed every 2 min https://dev-next-oblt.elastic.dev/app/apm#/services/test-apm-cli/transactions?rangeFrom=now-24h&rangeTo=now&transactionType=Pipeline

Finally, I have developed a pytest plugin to report APM traces from test runner with pytest.
image


vars/pipelineManager.groovy
```diff
@@ -41,6 +43,11 @@ def call(Map params = [:]) {
   if (firstTime) {
     log(level: 'INFO', text: 'firstTimeContributor step is not available yet.')
   }
+  if(apmTraces && isEnabled(apmTraces.get('when', 'always'))) {
```
v1v, line 46: `isEnabled` is case sensitive and uses `ALWAYS`

```suggestion
  if(isEnabled(apmTraces?.get('when', 'ALWAYS'))) {
```

PR #732, opened 2020-09-24
Fix nested when causes with aggregation

v1v:
## What does this PR do?

Allow nested when conditions with aggregation

## Why is it important?

Otherwise `macOs` builds are not enabled.

## Tasks

Need to think what's the best strategy from the configuration point of view.

## Related issues

Relates to https://github.com/elastic/beats/pull/21323


v1v:
I'll close this PR for the time being and revisit it in the future if we need to implement this feature

PR #735, opened 2020-09-28
Fix docker-compose and JCasC

v1v:
## What does this PR do?

Folder definition in JCasC should run firstly.
Docker compose reports an error:
```
services.jenkins.deploy.resources.reservations.cpus contains an invalid type, it should be a number
```

## Why is it important?

Otherwise, make -C local <command> won't work

```
$ make -C local stop 
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.jenkins.deploy.resources.reservations.cpus contains an invalid type, it should be a number
make: *** [stop] Error 1
```

### DIFF
--- a/local/configs/jenkins.yaml
+++ b/local/configs/jenkins.yaml
@@ -103,9 +103,9 @@ unclassified:
     globalConfigName: username
     globalConfigEmail: username@example.com
 jobs:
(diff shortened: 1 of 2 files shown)


mdelapenya (approved):

kuisathaverat:
I thought it was only my system 😸

PR #772, opened 2020-10-14
[notification] show only up to 10 tests/steps failures

v1v:
## What does this PR do?

When there are more than 10 tests failures or steps failures then show only the first 10 of each as a GitHub comment and add a quote comment regarding this filter

## Why is it important?

Avoid creating GitHub comments with thousands of lines.

For instance, 7.x branch got a few hundred of test failures:

image

and PRs got massive GitHub comments that were hard to debug, see https://github.com/elastic/beats/pull/21637#issuecomment-705027324

## Tests

UTs output

image

### DIFF
--- a/resources/github-comment-markdown.template
+++ b/resources/github-comment-markdown.template
@@ -54,7 +54,8 @@
   ### Test errors [![${testsSummary?.failed}](https://img.shields.io/badge/${testsSummary?.failed}%20-red)](${jobUrl}/tests)
   <details><summary>Expand to view the tests failures</summary>
   <p>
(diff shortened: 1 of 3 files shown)


mdelapenya (approved):

mdelapenya:
It would be great if each repo is able to configure this (and other CI-related stuff), like reading a YAML file with this values, maybe) 🤔

v1v:
> It would be great if each repo is able to configure this (and other CI-related stuff), like reading a YAML file with this values, maybe) 🤔

I like it, probably we need to find certain patterns, let me raise an issue with this

PR #773, opened 2020-10-15
[flaky] enable if no aborted

v1v:
## What does this PR do?

Builds that got aborted (normally a new build with new changes are in place) then no flaky test analyser should happen since there might not be tests to be analysed.

## Why is it important?

Avoid any kind of misleading 

For instance 

image

image

## Related issues
Closes #ISSUE

### DIFF
--- a/vars/notifyBuildResult.groovy
+++ b/vars/notifyBuildResult.groovy
@@ -77,8 +77,8 @@ def call(Map args = [:]) {
           }
         }
 
-        // Should analyze flakey
-        if(analyzeFlakey) {
+        // Should analyze flakey but exclude it when aborted
+        if(analyzeFlakey && currentBuild.currentResult != 'ABORTED') {
           data['es'] = es
           data['es_secret'] = secret
           data['flakyReportIdx'] = flakyReportIdx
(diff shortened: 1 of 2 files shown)


vars/notifyBuildResult.groovy
```diff
@@ -77,8 +77,8 @@ def call(Map args = [:]) {
           }
         }
 
-        // Should analyze flakey
-        if(analyzeFlakey) {
+        // Should analyze flakey but exclude it when aborted
+        if(analyzeFlakey && currentBuild.currentResult != 'ABORTED') {
```
mdelapenya, line 81: Is there a way to also check that the tests for the current build are all passing or empty?

mdelapenya (approved):

vars/notifyBuildResult.groovy
v1v: The idea was, if `analyzeFlakey` is enabled then the analysis of test results will happen even if all of them are passing there will be a notification. Otherwise, if for some reason the message is not reported we might not notice it straight away

Why do you ask for?

cachedout (approved):

vars/notifyBuildResult.groovy
v1v: One of the reasons to always notify independent of the test status is if there were a few commits with a few builds, then the last comment might not reflect the status of the last commit:
- For instance, if the last comment means a flaky failure and the last commit fixes then if the comment does not get changed then it won't reflect the right status.

Let's move this discussion to an issue if you think it's worthy to avoid notifying status when no test failures.

PR #827, opened 2020-11-09
[dependabot] disable docker images

v1v:
## What does this PR do?

Disable docker bump

## Why is it important?

As long as we don't run tests for the docker images we cannot warranty this is a valid use case


v1v:
@dependabot close

https://github.com/elastic/apm-pipeline-library/pull/827

kuisathaverat (approved):

cachedout (approved):

PR #847, opened 2020-11-16
Skip notifications if no notifications are in the list to send

v1v:
## Why is it important?

As a consequence of adding a new feature then notifications were not double checked and some notifications were sent when they should not.

Thanks @brianseeders


src/test/groovy/NotifyBuildResultStepTests.groovy
```diff
@@ -298,6 +299,7 @@ class NotifyBuildResultStepTests extends ApmBasePipelineTest {
   }
 
   @Test
+  @Ignore("Class does not have access to the built-in steps. Error: No signature of method: co.elastic.NotificationManager.catchError()")
```
v1v, line 302: See https://github.com/elastic/apm-pipeline-library/issues/848

mdelapenya (approved):

PR #855, opened 2020-11-19
Add certain latency to access to the google storage

v1v:
## What does this PR do?

Try to reduce the number of concurrency calls for the same step that causes some issues

## Why is it important?

See if we can mitigate the environmental issue as long as no new release of the plugins is released

I'm not 100% sure whether this particular workaround will reduce the number of times we face that issue, but it should be pretty much harmless in the worst case scenario, but adding certain sleep time.

## Related issues

Caused by https://github.com/jenkinsci/google-storage-plugin/issues/61

## Tests

Manually tried in my local instance:

image


mdelapenya (approved):

kuisathaverat (approved):
let's do a try

PR #856, opened 2020-11-20
Bulk update with a flatten tests

v1v:
## What does this PR do?

In order to keep working on dashboards is required to enable a new index with flatten test entries, for such, there will be a bulk update and every test will be a document rather than one document with nested maps.

## Why is it important?

Change the paradigm of document based rather than using the relational approach.

We will keep backward compatibility for the time being. 

## Questions

- Will the `bulk` work as expected in terms of new lines? since `\n` is the delimiter.

## Actions

- [x] Create a new index
- [x] Create the mapping to be effective

## What did I do?

- Create a index template:

```json
PUT _index_template/ci-builds-template
{
  "template": {
    "mappings": {
       ...
    "aliases": {
      "ci-builds": {}
    }
  },
  "index_patterns": [
    "ci-builds*"
  ]
}
```

- Create an index with the name `ci-builds-00001`

```bash
PUT /ci-builds-000001
```

- Enable write access to the alias

```json
POST /_aliases
{
    "actions" : [
        {
            "add" : {
                 "index" : "ci-builds-000001",
                 "alias" : "ci-builds",
                 "is_write_index" : true
            }
        }
    ]
}
```

- Grant bulk updates for the new index for the jenkins user


vars/notifyBuildResult.groovy
```diff
@@ -140,6 +140,11 @@ def call(Map args = [:]) {
         timeout(5) {
           def datafile = readFile(file: 'build-report.json')
           sendDataToElasticsearch(es: es, secret: secret, data: datafile)
+          def bulkFile = 'build-report-bulk.json'
+          if (fileExists(bulkFile)) {
+            datafile = readFile(file: bulkFile)
+            sendDataToElasticsearch(es: es, secret: secret, data: datafile, restCall: '/ci-builds/_bulk/')
```
v1v, line 146: This is the alias!

mdelapenya (approved):

cachedout (approved):
Nice!

resources/scripts/generate-build-data.sh
```diff
@@ -417,4 +418,13 @@ fetchAndPrepareBuildInfo "${BUILD_INFO}" "${BO_BUILD_URL}/" "build" "${DEFAULT_H
 prepareEnvInfo "${ENV_INFO}" "env"
 echo '}' >> "${BUILD_REPORT}"
 
+### For each entry in the test map then create a flatten document
+N=0
+jq -c '.test = (.test[])' "${BUILD_REPORT}" |
+while read -r json ; do
+  N=$((N+1))
+  echo "{ \"index\":{} }" >> "${BUILD_BULK_REPORT}"
+  echo "${json}" >> "${BUILD_BULK_REPORT}"
```
kuisathaverat, line 427:
```suggestion
  echo "{ \"doc\" : ${json} }" >> "${BUILD_BULK_REPORT}"
```

v1v: Cool, I just added some tests to ensure the `ndjson` format works as expected, and in fact `-c` is the key flag for the jq

PR #857, opened 2020-11-23
🚧 [CI] Automate bump versions

v1v:
## What does this PR do?

Automate the process for bumping the versions.

## Why is it important?

Avoid manual actions for something that happens with certain frequency .

## Tasks
- [ ] Define what projects could benefit of this.
- [ ] Define the contracts with all those projects. (for instance `.ci/scripts/bump_version.sh` and `.ci/scripts/return_latest_version.sh` should be the scripts created in each repo)

## Issues

Depends on https://github.com/elastic/observability-test-environments/pull/937

### DIFF
--- a/.ci/schedule-daily.groovy
+++ b/.ci/schedule-daily.groovy
@@ -24,6 +24,7 @@ pipeline {
     PIPELINE_LOG_LEVEL='INFO'
     DOCKERHUB_SECRET = 'secret/apm-team/ci/elastic-observability-dockerhub'
     DOCKERELASTIC_SECRET = 'secret/apm-team/ci/docker-registry/prod'
+    JOB_GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
   }
   options {
     timeout(time: 1, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


.ci/schedule-daily.groovy
```diff
@@ -88,6 +89,54 @@ pipeline {
...
+            steps {
+              gitCheckout(basedir: '.', branch: env.BRANCH,
+                repo: 'git@github.com:elastic/observability-test-environments.git',
+                credentialsId: "${JOB_GIT_CREDENTIALS}",
+                githubNotifyFirstTimeContributor: false)
+            }
+          }
+          stage('Fetch latest version'){
+            steps {
+              echo 'TBD - Search the latest version'
```
kuisathaverat, line 115: my manual process for this 😄 
* Check https://artifacts-api.elastic.co/v1/versions/
* Pick the next [major|minor|patch] snapshot/BC

so we can run it on schedule something like this:
* looking for the next patch [snapshot|BC]
  * if there is not minor check for the next minor
  * if there is not minor check for the next major.
* Check if there is a PR for the update found
  * if there is no PR Create a PR for the update else skip the PR creation

v1v:
this is now obsoleted, see https://github.com/elastic/apm-pipeline-library/commit/951ac820d02e0851766cdaab33acd6db6016b2ed

PR #862, opened 2020-11-25
[ignore] gitcmd files with the .git folder

v1v:
## What does this PR do?

Store logs in the .git folder when possible

## Why is it important?

To avoid adding dirty files in the git repo


mdelapenya (approved):
Great!!

PR #864, opened 2020-11-25
[index] populate only test failures

v1v:
## What does this PR do?

Reduce the ndjson files for the bulk from 100mbs in Beats to a few kbs in the new indexes, so we won't loose any data yet since we still have the deprecated index with a multinested doc fields.

## Why is it important?

Need to sacrifice those details otherwise the 100mbs data causes issues with the `_bulk` entrypoint.

image

### DIFF
--- a/resources/ci-tests-mapping.json
+++ b/resources/ci-tests-mapping.json
@@ -257,6 +257,14 @@
               }
             }
           },
+          "errorDetails" : {
+            "type" : "keyword",
+            "fields" : {
+              "analyzed" : {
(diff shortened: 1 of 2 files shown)


resources/ci-tests-mapping.json
```diff
@@ -257,6 +257,14 @@
               }
             }
           },
+          "errorDetails" : {
```
v1v, line 260: leftover

kuisathaverat (approved):

PR #868, opened 2020-12-02
[daily] build docker images

v1v:
## What does this PR do?

Enable rum-js, apm-integration and apm-server docker image generation on a daily basis

## Why is it important?

Cache as much as possible


.ci/schedule-daily.groovy
```diff
@@ -52,14 +52,16 @@ pipeline {
...
             string(name: 'secret', value: 'secret/apm-team/ci/docker-registry/prod'),
+            booleanParam(name: 'apm_integration_testing', value: true),
+            booleanParam(name: 'apm_server', value: true),
+            booleanParam(name: 'helm_kubectl', value: true),
             booleanParam(name: 'nodejs', value: true),
+            booleanParam(name: 'opbot', value: true),
+            booleanParam(name: 'oracle_instant_client', value: true),
             booleanParam(name: 'python', value: true),
             booleanParam(name: 'ruby', value: true),
+            booleanParam(name: 'rum', value: true),
```
mdelapenya, line 63: What is the difference with https://github.com/elastic/apm-pipeline-library/blob/master/.ci/buildDockerImages.groovy, which is scheduled daily?

v1v: I don't see any trigger there, besides the pipeline got a default `false` parameters, so even if it gets triggered it won't do anything.

`buildDockerImages.groovy` is only the how while `schedule-daily.groovy` is the `when`

mdelapenya (approved):
thanks for the clarifications!

PR #884, opened 2021-01-04
feat: new goTestJunit step

kuisathaverat:
## What does this PR do?

A new step to run Go Unit tests and generate a JUnit report.

## Why is it important?

It simplifies/unifies the Go test executions across projects.


cachedout (approved):

vars/goTestJunit.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+ goTestJunit(options: '-v ./...', output: 'build/junit-report.xml')
+
+*/
+def call(Map args = [:]) {
+  def options = args.containsKey('options') ? args.options : ''
+  def output = args.containsKey('output') ? args.output : 'junit-report.xml'
+  log(level: 'INFO', text: 'Running Go test an generating JUnit output' + text)
+
+  withGoEnv(pkgs: ["github.com/jstemmer/go-junit-report"]){
```
mdelapenya, line 29: Should the package be `gotest.tools/gotestsum`?
```suggestion
  withGoEnv(pkgs: ["gotest.tools/gotestsum"]){
```

vars/goTestJunit.groovy
```diff
@@ -0,0 +1,32 @@
...
+    cmd(label: 'Running Go tests' , script: "gotestsum --junitfile ${output} -- -v ${options}")
```
mdelapenya, line 30: Also, to improve the output in Jenkins logs, if we set `--format testname` instead, and we set an env var `GOTESTSUM_JUNITFILE=${output}` then we have both things at the same time: good log and jUnit file.

kuisathaverat: I prefer to use only options or only env vars

mdelapenya (approved):

vars/withGoEnv.groovy
```diff
@@ -44,3 +45,9 @@ def call(Map args = [:], Closure body) {
     }
   }
 }
+
+def checkGoPath(){
+  if(fileExists(file: "${env.GOPATH}/go.mod")){
```
v1v, line 50: is `GOPATH` defined earlier? If so, where?

vars/goTestJunit.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+ goTestJUnit(options: '-v ./...', output: 'build/junit-report.xml')
+
+*/
+def call(Map args = [:]) {
+  def options = args.containsKey('options') ? args.options : ''
+  def output = args.containsKey('output') ? args.output : 'junit-report.xml'
+  log(level: 'INFO', text: 'Running Go test an generating JUnit output')
+
+  withGoEnv(pkgs: ["gotest.tools/gotestsum"]){
```
v1v, line 29: I wonder if the `go-version` should be also somehow customised or enabled to be customised?

vars/withGoEnv.groovy
kuisathaverat: it is defined by `withGoEnv`

vars/goTestJunit.groovy
kuisathaverat: it is you can use the GO_VERSION environment variable.

kuisathaverat: in any case, I've added it.

vars/withGoEnv.groovy
v1v: right, but it does not exist yet when the checkGoPath is called

v1v (changes requested):
See my comment regarding the `checkGoPath` method

vars/withGoEnv.groovy
kuisathaverat: I've moved the call inside the `withGoEnvUnix` and `withGoEnvWindows`  closures

v1v (approved):

PR #885, opened 2021-01-04
chore: rename argument

kuisathaverat:
## What does this PR do?

Rename the argument `params` to `args`in the dummy step.

## Why is it important?

This step is the template for new steps, the use of `args` as name for the arguments is better and avoid conflicts/confusion with the global `params` variable.

## Related issues
Closes #ISSUE

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 .idea
 *.iml
 release.properties
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.DS_Store
 .idea
 *.iml
 release.properties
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
v1v, line 1: NIT: if you configure your global `${HOME}/.gitignore` then this won't be required

```
$ cat ~/.gitignore        
...
# Folder view configuration files
.DS_Store
Desktop.ini
...
```

v1v (approved):

cachedout (approved):

.gitignore
kuisathaverat, line 1:
```suggestion
```

kuisathaverat: you are right.

```
echo ".DS_Store" >> ~/.gitignore  
git config --global core.excludesfile ~/.gitignore
```

PR #890, opened 2021-01-08
Support not_changeset_full_match for beats

v1v:
## What does this PR do?

Support a new when condition to full match the changeset

## Why is it important?

Otherwise, there is a chance that some PRs won't trigger mandatory fields since it uses a partial match at the moment.

## For the review

- First commit (https://github.com/elastic/apm-pipeline-library/pull/890/commits/e1ba36d22b1b1dc09b5a0b795becf2f7f9b11db4) was a refactor, so almost all the logic for the `whenChangeset` was moved to a new function `changeset`.
- Second commit (https://github.com/elastic/apm-pipeline-library/pull/890/commits/578f4f80b197a9d6a9b21098ecb3c3e5f2d9d461), added the new when condition.

### DIFF
--- a/vars/beatsWhen.groovy
+++ b/vars/beatsWhen.groovy
@@ -64,7 +64,7 @@ private Boolean whenChangeset(Map args = [:]) {
   if (args.content?.get('changeset')) {
     def arguments = args
     arguments.name = name
-    return changeset(args)
+    return changeset(arguments)
   } else {
     markdownReason(project: args.project, reason: "* ${name} is `disabled`.")
   }
--- a/vars/beatsWhen.groovy
+++ b/vars/beatsWhen.groovy
@@ -38,6 +38,7 @@ Boolean call(Map args = [:]){
     if (whenComments(args)) { ret = true }
     if (whenLabels(args)) { ret = true }
     if (whenNotChangeset(args)) { ret = true }
+    if (whenNotChangesetFullMatch(args)) { ret = true }
     if (whenParameters(args)) { ret = true }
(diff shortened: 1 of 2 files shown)


vars/beatsWhen.groovy
```diff
@@ -59,63 +60,81 @@ private Boolean whenBranches(Map args = [:]) {
...
 private Boolean whenChangeset(Map args = [:]) {
+  def name = 'Changeset'
   if (args.content?.get('changeset')) {
-    // Gather macro changeset entries
-    def macro = [:]
-    args?.changeset?.each { k,v ->
-      macro[k] = v
-    }
+    def arguments = args
+    arguments.name = name
```
cachedout, line 66: I'm not understanding the need for these two lines. Am I just not seeing where `arguments` is used?

cachedout (approved):

vars/beatsWhen.groovy
v1v: aha! good catch

vars/beatsWhen.groovy
```diff
@@ -59,63 +60,81 @@ private Boolean whenBranches(Map args = [:]) {
...
+    return changeset(args)
```
v1v, line 67:
```suggestion
    return changeset(arguments)
```

vars/beatsWhen.groovy
v1v: I've just fixed below, but I'll add some UTs now

kuisathaverat (approved):

PR #894, opened 2021-01-11
[CI][mvn4] Use lastSuccessfulBuild

v1v:
## What does this PR do?

Use link to the latest successfull build instead a specific build

## Why is it important?

Avoid using old deleted builds.

It's not ideal but as long as mvn 4 is not released in a different manner, we cannot do much unless we archive those artifacts somewhere.

## Important

This will not add reproducibility to our builds since it'll use the latest stable artifacts, but I think once we are on mvn4 any incremental changes should be backward compatibly, if not, at least we can provide value feedback to the maven project.

I'll ask the maven project if they have another suggestion to use their latest mvn 4 artifacts


kuisathaverat (approved):

.mvn/wrapper/maven-wrapper.properties
```diff
@@ -1,2 +1,2 @@
-distributionUrl=https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven/job/master/70/artifact/org/apache/maven/apache-maven/4.0.0-alpha-1-SNAPSHOT/apache-maven-4.0.0-alpha-1-SNAPSHOT-bin.zip
+distributionUrl=https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven/job/master/lastSuccessfulBuild/artifact/org/apache/maven/apache-maven/4.0.0-alpha-1-SNAPSHOT/apache-maven-4.0.0-alpha-1-SNAPSHOT-bin.zip
```
mdelapenya, line 1: Neat!

mdelapenya (approved):
Fixed, cool!

image

.mvn/wrapper/maven-wrapper.properties
v1v: It's a temporary change, I need to find a better way to make our builds reproducible

PR #896, opened 2021-01-12
chore: use node name as json file trying to be more deterministic

mdelapenya:
## What does this PR do?
It uses node name, obtained from Jenkins' NODE_NAME variable, as file name for the JSON file generated to store the state of the filebeat container.

## Why is it important?
We are seeing situations where the ID stored in the file is not relating to the filebeat container, causing errors when trying to stop the container.

We already know that the JSON files managed by `writeJSON` and `readJSON` helper methods are directly stored in the worker, not in the master, but this change adds an unique file name for the JSON.

## Related issues
Closes #ISSUE


v1v (approved):

src/test/groovy/FilebeatStepTests.groovy
```diff
@@ -22,11 +22,14 @@ import static org.junit.Assert.assertFalse
 
 class FilebeatStepTests extends ApmBasePipelineTest {
   String scriptName = 'vars/filebeat.groovy'
+  // test resources file uses this value as filename
+  String nodeName = 'worker-0676d01d9601f8191'
```
cachedout, line 26: It appears that you are always using this to compose the same string, which is `filebeat_container_worker-0676d01d9601f8191.json`. Would it be better to just define it once and re-use it?

cachedout (approved):

kuisathaverat (approved):

src/test/groovy/FilebeatStepTests.groovy
mdelapenya: Done!

Issue #897, opened 2021-01-12
CI complains about dir() and filebeat()


mtojek:
I did some research and found out that whenever `filebeat()` is included within any `dir()` it fails:

Sample:

```groovy
                  node('ubuntu && immutable') {
                    stage("${it}: check") {
                      dir("${env.NODE_NAME}") {
                        deleteDir()
                        gitCheckout(basedir: "${BASE_DIR}")
                        useElasticPackage()
                        try {
                          filebeat() {
                            dir("${BASE_DIR}/packages/${it}") {
                              sh(label: "Check integration: ${it}", script: '../../build/elastic-package check -v')
                              // Check if there are test folders present, so the CI job runner can spawn the Elastic stack
                              // locally and run tests for the integration. The condition prevents from spawning the stack
                              // unnecessarily (while no tests are defined).
                              def r = sh(label: "Find test folders", script: 'find . -d  -name test | grep .', returnStatus: true)
                              if (r == 0) {
                                sh(label: "Boot up the Elastic stack", script: '../../build/elastic-package stack up -d -v')
                                sh(label: "Test integration: ${it}", script: '''
                                  eval "$(../../build/elastic-package stack shellinit)"
                                  ../../build/elastic-package test -v --report-format xUnit --report-output file
                                  ''')
                                sh(label: "docker ps", script: 'docker ps -a')
                              }
                            }
                          }
```

For reference: https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Fintegrations/detail/PR-513/13/pipeline

Output:

```
[2021-01-12T14:05:14.543Z] + docker run --detach -v /var/lib/jenkins/workspace/gest-manager_integrations_PR-513:/output -v /var/lib/jenkins/workspace/gest-manager_integrations_PR-513/filebeat_conf.yml:/usr/share/filebeat/filebeat.yml -u 0:0 -v /var/lib/docker/containers:/var/lib/docker/containers -v /var/run/docker.sock:/var/run/docker.sock docker.elastic.co/beats/filebeat:7.10.1 --strict.perms=false -environment container -E http.enabled=true
[2021-01-12T14:05:14.543Z] Unable to find image 'docker.elastic.co/beats/filebeat:7.10.1' locally
[2021-01-12T14:05:15.483Z] 7.10.1: Pulling from beats/filebeat
[2021-01-12T14:05:15.483Z] a4f595742a5b: Pulling fs layer
[...]

PR #900, opened 2021-01-13
fix: use current folder instead of the workspace

kuisathaverat:
## What does this PR do?

Use the current dir by default for all generated files.

## Why is it important?

ArchiveArtifacts uses relative routes to the current folder, so grab files outside this scope make the step complex without need, because of that we change to use the current folder as workdir for the generated files.

## Related issues
Closes https://github.com/elastic/apm-pipeline-library/issues/897

### DIFF
--- a/vars/filebeat.txt
+++ b/vars/filebeat.txt
@@ -2,7 +2,7 @@
  This step run a filebeat Docker container to grab the Docker containers logs in a single file.
  `filebeat.stop()` will stop the Filebeat Docker container and grab the output files,
  the only argument need is the `workdir` if you set it on the `filebeat step` call.
- The output log files should be in a relative parth to the current path (see [archiveArtifacts](https://www.jenkins.io/doc/pipeline/steps/core/#archiveartifacts-archive-the-artifacts))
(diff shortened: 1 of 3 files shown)


vars/filebeat.txt
```diff
@@ -2,6 +2,7 @@
  This step run a filebeat Docker container to grab the Docker containers logs in a single file.
  `filebeat.stop()` will stop the Filebeat Docker container and grab the output files,
  the only argument need is the `workdir` if you set it on the `filebeat step` call.
+ The output log files should be in a relative parth to the current path (see [archiveArtifacts](https://www.jenkins.io/doc/pipeline/steps/core/#archiveartifacts-archive-the-artifacts))
```
mdelapenya, line 5:
```suggestion
 The output log files should be in a relative path to the current path (see [archiveArtifacts](https://www.jenkins.io/doc/pipeline/steps/core/#archiveartifacts-archive-the-artifacts))
```

mdelapenya (approved):

cachedout (approved):

PR #901, opened 2021-01-13
feat: goDefaultVersion step

kuisathaverat:
## What does this PR do?

Add a new step to get the default Go version we use.

## Why is it important?

We have duplicate logic across projects to do exactly the same, also we have harcoded values we have to update manually across repos, this resolves both issues.

## Related issues
caused by https://github.com/elastic/ecs-logging-go-zap/pull/28

### DIFF
--- a/vars/goDefaultVersion.groovy
+++ b/vars/goDefaultVersion.groovy
@@ -1,0 +1,34 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


vars/goDefaultVersion.groovy
```diff
@@ -0,0 +1,34 @@
...
+
+  goDefaultVersion()
+**/
+def call(Map args = [:]) {
+  def goDefaultVersion = isGoVersionEnvVarSet() ? "${env.GO_VERSION}" : '1.15.6'
+  def goFileVersion = '.go-version'
+  if(!isGoVersionEnvVarSet() && fileExists(goFileVersion)){
+    goDefaultVersion = readFile(file: goFileVersion)?.trim()
+  }
+  return  goDefaultVersion
```
v1v, line 29:
```suggestion
  def goDefaultVersion = '1.15.6'
  if(isGoVersionEnvVarSet()) {
    goDefaultVersion = "${env.GO_VERSION}"
  } else {
    def goFileVersion = '.go-version'
    if (fileExists(goFileVersion)){
      goDefaultVersion = readFile(file: goFileVersion)?.trim()
    }
  }
  return  goDefaultVersion
```

v1v (approved):
NIT: with a more explicit conditions

PR #902, opened 2021-01-13
feat: add stackVersions step

kuisathaverat:
## What does this PR do?

It adds a new step stackVersions.

## Why is it important?

This new step returns an array of the version numbers we use for testing, also can give the version using tag names (edge, dev, release, or snapshot)

## Related issues
related to https://github.com/elastic/observability-robots/issues/309

### DIFF
--- a/vars/stackVersions.groovy
+++ b/vars/stackVersions.groovy
@@ -27,11 +27,11 @@
   stackVersions.edge(snapshot: true)
 
 **/
-def call(Map params = [:]) {
+def call(Map args = [:]) {
   return [
-    edge(params),
-    dev(params),
-    release(params)
+    edge(args),
+    dev(args),
+    release(args)
   ]
 }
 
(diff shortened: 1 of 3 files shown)


vars/stackVersions.groovy
```diff
@@ -0,0 +1,60 @@
...
+  return version("7.11.0", params)
+}
+
+def release(Map params = [:]){
+  return version("7.10.2", params)
+}
+
+def isSnapshot(params){
+  return params.containsKey('snapshot') && params.snapshot ? "-SNAPSHOT" : ''
+}
```
v1v, line 60:
```suggestion
def call(Map args = [:]) {
  return [
    edge(args),
    dev(args),
    release(args)
  ]
}

def snapshot(version){
  return "${version}-SNAPSHOT"
}

def version(value, args = [:]){
  return "${value}${isSnapshot(args)}"
}

def edge(Map args = [:]){
  return version("8.0.0", args)
}

def dev(Map args = [:]){
  return version("7.11.0", args)
}

def release(Map args = [:]){
  return version("7.10.2", args)
}

def isSnapshot(args){
  return args.containsKey('snapshot') && args.snapshot ? "-SNAPSHOT" : ''
}
```

v1v (approved):

PR #903, opened 2021-01-13
feat: use goDefaultVersion

kuisathaverat:
## What does this PR do?

It changes some steps to use the new goDefaultVersion step.

## Why is it important?

This unifies the Go version used across steps and avoid duplications.

## Related issues
related to https://github.com/elastic/apm-pipeline-library/pull/901

### DIFF
--- a/src/test/groovy/GoTestJUnitStepTests.groovy
+++ b/src/test/groovy/GoTestJUnitStepTests.groovy
@@ -34,11 +34,13 @@ class GoTestJUnitStepTests extends ApmBasePipelineTest {
 
   @Test
   void test() throws Exception {
+    def scriptDV = loadScript('vars/goDefaultVersion.groovy')
+    def version = scriptDV.defaultVersion()
     def script = loadScript(scriptName)
     script.call()
     printCallStack()
     assertTrue(assertMethodCallContainsPattern('sh', 'gotestsum --format testname --junitfile junit-report.xml --'))
-    assertTrue(assertMethodCallContainsPattern('withGoEnvUnix', 'version=null'))
+    assertTrue(assertMethodCallContainsPattern('withGoEnvUnix', "version=${version}"))
     assertJobStatusSuccess()
   }
 
--- a/src/test/groovy/GoTestJUnitStepTests.groovy
+++ b/src/test/groovy/GoTestJUnitStepTests.groovy
@@ -34,8 +34,8 @@ class GoTestJUnitStepTests extends ApmBasePipelineTest {
 
   @Test
   void test() throws Exception {
-    def scriptDV = loadScript('vars/goDefaultVersion.groovy')
-    def version = scriptDV.defaultVersion()
+    def version = "1.15.1"
+    helper.registerAllowedMethod('goDefaultVersion', [], { version })
     def script = loadScript(scriptName)
     script.call()
     printCallStack()
--- a/src/test/groovy/WithGoEnvUnixStepTests.groovy
+++ b/src/test/groovy/WithGoEnvUnixStepTests.groovy
@@ -122,8 +122,8 @@ void testOSArg() throws Exception {
   @Test
   void testDefaultGoVersion() throws Exception {
     helper.registerAllowedMethod('nodeOS', [], { "linux" })
-    def scriptDV = loadScript('vars/goDefaultVersion.groovy')
-    def version = scriptDV.defaultVersion()
+    def version = "1.15.1"
+    helper.registerAllowedMethod('goDefaultVersion', [], { version })
     def isOK = false
     script.call(){
       if(binding.getVariable("PATH") == "WS/bin:WS/.gvm/versions/go${version}.linux.amd64/bin:/foo/bin"
--- a/src/test/groovy/WithGoEnvUnixStepTests.groovy
+++ b/src/test/groovy/WithGoEnvUnixStepTests.groovy
@@ -122,17 +122,19 @@ void testOSArg() throws Exception {
   @Test
   void testDefaultGoVersion() throws Exception {
     helper.registerAllowedMethod('nodeOS', [], { "linux" })
+    def scriptDV = loadScript('vars/goDefaultVersion.groovy')
+    def version = scriptDV.defaultVersion()
     def isOK = false
     script.call(){
-      if(binding.getVariable("PATH") == "WS/bin:WS/.gvm/versions/go1.14.2.linux.amd64/bin:/foo/bin"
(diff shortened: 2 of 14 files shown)


src/test/groovy/WithGoEnvUnixStepTests.groovy
```diff
@@ -122,17 +122,19 @@ void testOSArg() throws Exception {
   @Test
   void testDefaultGoVersion() throws Exception {
     helper.registerAllowedMethod('nodeOS', [], { "linux" })
+    def scriptDV = loadScript('vars/goDefaultVersion.groovy')
+    def version = scriptDV.defaultVersion()
```
v1v, line 126: Uhm, when I used `step`.method in the past, I used to use a mock class and then use the `helper.property` method, see https://github.com/elastic/apm-pipeline-library/blob/85bc3987206c6aad98744a576c6e61e8692374be/src/test/groovy/ApmBasePipelineTest.groovy#L82-L86

What do you think?

vars/sendBenchmarks.groovy
```diff
@@ -53,53 +53,52 @@ def call(Map params = [:]) {
...
-          }
-      }
-   }
+  if(index.equals('benchmark-go') || index.equals('benchmark-server')){
+    def protocol = getProtocol(url)
+    url = url - protocol
+    url = "${protocol}${user}:${password}@${url}"
+    gobench(url, benchFile, index)
+  } else {
+    cloudBecnhmarks(url, user, password, benchFile)
```
v1v, line 64:
```suggestion
    cloudBenchmarks(url, user, password, benchFile)
```

vars/sendBenchmarks.groovy
```diff
@@ -53,53 +53,52 @@ def call(Map params = [:]) {
...
+  }
+}
+
+def cloudBecnhmarks(realURL, user, password, benchFile) {
```
v1v, line 68:
```suggestion
def cloudBenchmarks(realURL, user, password, benchFile) {
```

vars/sendBenchmarks.groovy
```diff
@@ -53,53 +53,52 @@ def call(Map params = [:]) {
...
-              error "Benchmarks: there was a response with an error. Review response ${response}"
-            }
-          }
-      }
-   }
+  if(index.equals('benchmark-go') || index.equals('benchmark-server')){
+    def protocol = getProtocol(url)
+    url = url - protocol
+    url = "${protocol}${user}:${password}@${url}"
+    gobench(url, benchFile, index)
```
v1v, line 62:
```suggestion
    gobench("${protocol}${user}:${password}@${url}", benchFile, index)
```

src/test/groovy/WithGoEnvUnixStepTests.groovy
kuisathaverat: the use of this step is to do not have to modify the test every time we change the default Go version, if we use a mock we have to change the version in the step and in the mock.

v1v: Maybe `defaultVersion` should be an unique step, renamed to `staticGoDefaultVersion` and then it will be easy to be mocked in the context of this particular class with:

```
    helper.registerAllowedMethod('staticGoDefaultVersion', [], { "1.15.5" })
    if(binding.getVariable("PATH") == "WS/bin:WS/.gvm/versions/go1.15.5.linux.amd64/bin:/foo/bin"
       && binding.getVariable("GOROOT") == "WS/.gvm/versions/go1.15.5.linux.amd64"
...
```

vars/goDefaultVersion.groovy
```diff
@@ -36,3 +36,7 @@ def call(Map args = [:]) {
 def isGoVersionEnvVarSet(){
   return env.GO_VERSION != null && "" != "${env.GO_VERSION}"
 }
+
+def defaultVersion(){
+  return '1.15.6'
```
v1v, line 41:
```suggestion
  return staticGoDefaultVersion()
```

kuisathaverat: I do not get this.

src/test/groovy/WithGoEnvUnixStepTests.groovy
kuisathaverat: I am not sure you can mock a method from the same class in that way. I will check.

vars/goDefaultVersion.groovy
v1v: `step.method` cannot be mocked then it requires to use in the UTs the below snippet:

```
   def script = loadScript('vars/step.groovy')
   def value = script.method()
```

As far as I see `step.method` in our case is not used but only for given dynamism when running the UTs.

If that's the case, I was suggesting:

* create a new step called `staticGoDefaultVersion`
* then mock the step`helper.registerAllowedMethod('staticGoDefaultVersion', ...` in WithGoEnvUnixStepTests

src/test/groovy/WithGoEnvUnixStepTests.groovy
kuisathaverat: I realized that I can mock the function directly

vars/goDefaultVersion.groovy
kuisathaverat: finally was simpler https://github.com/elastic/apm-pipeline-library/pull/903#discussion_r557517613

src/test/groovy/GoTestJUnitStepTests.groovy
```diff
@@ -34,11 +34,13 @@ class GoTestJUnitStepTests extends ApmBasePipelineTest {
 
   @Test
   void test() throws Exception {
+    def scriptDV = loadScript('vars/goDefaultVersion.groovy')
+    def version = scriptDV.defaultVersion()
```
v1v, line 38:
```suggestion
    def version = "1.15.1"
    helper.registerAllowedMethod('goDefaultVersion', [], { version })
```

src/test/groovy/WithGoEnvWindowsStepTests.groovy
```diff
@@ -109,13 +109,15 @@ void testOSArg() throws Exception {
   @Test
   void testDefaultGoVersion() throws Exception {
     helper.registerAllowedMethod('nodeOS', [], { "windows" })
+    def scriptDV = loadScript('vars/goDefaultVersion.groovy')
+    def version = scriptDV.defaultVersion()
```
v1v, line 113:
```suggestion
    def version = "1.15.1"
    helper.registerAllowedMethod('goDefaultVersion', [], { version })
```

v1v (approved):

PR #935, opened 2021-01-29
WIP [Usability] Access dynamically to the stage BO URL linked to the GH Check

v1v:
## What does this PR do?

Trying to support GitHub check URLs that point to the Blue Ocean Stage URL with the logs.

## Why is it important?

Avoid any further access to anything that's not BlueOcean

### DIFF
--- a/vars/withGithubCheck.groovy
+++ b/vars/withGithubCheck.groovy
@@ -15,6 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
+import org.jenkinsci.plugins.workflow.graph.StepNode
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor
+
 /**
   Wrap the GitHub notify check step
 
@@ -38,6 +41,13 @@ def call(Map args = [:], Closure body) {
   def isBo = args.get('isBlueOcean', false)
 
   if (!redirect.startsWith('http')) {
+
+    // Let's now support a link to the stage logs
+    if (redirect.equals('pipeline')) {
+      redirect = getStageLogUrl()
+    }
+
+    // Get the URL for the given tab.
     if (isBo) {
       redirect = getBlueoceanTabURL(redirect)
     } else {
@@ -68,4 +78,24 @@ def call(Map args = [:], Closure body) {
 
 boolean isAvailable(Map args = [:]) {
   return (args.get('org', env.ORG_NAME) && args.get('repository', env.REPO_NAME) && args.get('commitId', env.GIT_BASE_COMMIT))
+}
+
+def getStageLogUrl(flowNode = null) {
+  if(!flowNode) {
+    flowNode = getContext(org.jenkinsci.plugins.workflow.graph.FlowNode)
+  }
+  if(isStageNode(flowNode)) {
+    return "/${flowNode.url}log/?start=0"
+  }
+
+  return flowNode?.parents?.findResult { getNodeWsUrl(it) }
+}
+
+private boolean isStageNode(node = null) {
+  if (node instanceof StepNode) {
+    StepDescriptor d = ((StepNode) node).getDescriptor()
+    return d != null && d.getFunctionName().equals("stage")
+  } else {
+    return false
+  }
 }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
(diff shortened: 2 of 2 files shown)


vars/getCurrentStage.groovy
```diff
@@ -0,0 +1,50 @@
...
+import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode
+import org.jenkinsci.plugins.workflow.actions.LabelAction
+
+/**
+  Get the stage object in the current context.
+
+  Returns https://javadoc.jenkins.io/plugin/workflow-cps/org/jenkinsci/plugins/workflow/cps/nodes/StepStartNode.html
+*/
+
+def call() {
```
v1v, line 27: Need to be tested in the CI

vars/withGithubCheck.groovy
```diff
@@ -66,6 +66,50 @@ def call(Map args = [:], Closure body) {
   }
 }
 
+String detailsUrl(redirect, isBo=false) {
```
v1v, line 69: This should be provided as isolated step then others can consume this, such as the withGithubStatus.

vars/detailsURL.groovy
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+Generate the details URL to be added to the GitHub notifications. When possible it will look for the stage logs URL in BlueOcean.
+
+  def url = detailsURL(tab: 'artifacts', isBlueOcean: true)
+*/
+def call(Map args = [:]) {
+  def tab = args.get('tab', 'pipeline')
+  def isBlueOcean = args.get('isBlueOcean', false)
+  // Use the https
```
mdelapenya, line 26: Is this comment misleading to `https` instead? I'd elaborate the reason for why using http instead
```suggestion
  // Use the http
```

vars/getBlueoceanRestURLJob.groovy
```diff
@@ -0,0 +1,34 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
mdelapenya, line 1: Super super NIT: wdyt about renaming this file (and tests and docs) to `getBlueOceanRestURLJob.groovy`, to be consistent with `Blue Ocean` camel case?

vars/withGithubCheck.groovy
```diff
@@ -34,16 +34,7 @@ def call(Map args = [:], Closure body) {
...
-  def isBo = args.get('isBlueOcean', false)
-
-  if (!redirect.startsWith('http')) {
-    if (isBo) {
-      redirect = getBlueoceanTabURL(redirect)
-    } else {
-      redirect = getTraditionalPageURL(redirect)
-    }
-  }
+  def redirect = detailsURL(tab: args.get('tab', 'pipeline'), isBlueOcean: args.get('isBlueOcean', false))
```
mdelapenya, line 37: I like having the input parameters list at the beginning of the method, so that it's possible to understand the expected params in the args map. Putting them here could hide them from an ocasional reader

vars/withGithubStatus.groovy
```diff
@@ -29,16 +29,9 @@
...
-  def isBo = args.get('isBlueOcean', false)
 
-  if (!redirect.startsWith('http')) {
-    if (isBo) {
-      redirect = getBlueoceanTabURL(redirect)
-    } else {
-      redirect = getTraditionalPageURL(redirect)
-    }
-  }
+  def redirect = detailsURL(tab: args.get('tab', 'pipeline'), isBlueOcean: args.get('isBlueOcean', false))
```
mdelapenya, line 33: Ditto

vars/detailsURL.groovy
v1v:
```suggestion
```

vars/getBlueoceanRestURLJob.groovy
v1v: Unfortunately we got an old step -> `getBlueoceanDisplayURL` so I'd prefer to keep both with the same format, for backward compatibility, Though `getBlueoceanDisplayURL` might not be used inside any declarative pipeline

kuisathaverat (approved):

vars/getBlueoceanRestURLJob.groovy
mdelapenya: Good to know, thanks for the explanation

mdelapenya (approved):

cachedout (approved):

PR #937, opened 2021-02-01
Enable opentelemetry locally

v1v:
## What does this PR do?

Support open-telemetry for our local development.

You can use `make -C local build start-all` then you can start building jobs and go to http://localhost:5601/app/apm/services/jenkins/transactions to see those transactions

## Why is it important?

Test https://github.com/cyrille-leclerc/opentelemetry-plugin

## Tasks

- [x] Configure docker for open-telemetry since It works for https://github.com/cyrille-leclerc/opentelemetry-plugin/blob/master/opentelemetry-collector/run-opentelemetry-collector.sh but it does not work for the container. Somehow the `elastic` exporter is not recognised

## Tests

Blue Ocean running locally

image

Caused the below APM transactions (with INFO log level, aka less number of echo steps)

transactions_without_debug_echo


kuisathaverat (approved):

local/configs/jenkins.yaml
```diff
@@ -102,6 +102,10 @@ unclassified:
   gitscm:
     globalConfigName: username
     globalConfigEmail: username@example.com
+
+  openTelemetry:
+    endpoint: "localhost:4317"
+    useTls: false
```
v1v, line 108: quick and dirty and might require the plugin to be installed beforehand

local/configs/plugins.txt
```diff
@@ -1,4 +1,5 @@
 disable-github-multibranch-status
 filesystem_scm
 job-dsl
-ssh-agent
\ No newline at end of file
+opentelemetry::https://storage.googleapis.com/apm-ci-temp/plugins/opentelemetry.hpi
```
v1v, line 4: I uploaded the plugin after building it locally

cyrille-leclerc:
Interesting, should we ignore some steps like the `echo` step?

v1v:
> Interesting, should we ignore some steps like the echo step?

@cyrille-leclerc , this particular project got a DEBUB log level, so that's the reason we see all those echo steps. Probably `echo` is not something to worry about unless there is a bug with some interpolation then I guess it might be useful to be trapped, what do you think?

cyrille-leclerc:
> > Interesting, should we ignore some steps like the echo step?
> 
> @cyrille-leclerc , this particular project got a DEBUB log level, so that's the reason we see all those echo steps. Probably `echo` is not something to worry about unless there is a bug with some interpolation then I guess it might be useful to be trapped, what do you think?

Sounds good to me, just an idea to keep in mind then.

cachedout (approved):

mdelapenya (approved):

cyrille-leclerc:
FYI the plugin won't create spans for some steps that are not very interesting to monitor:"dir", "echo", "isUnix", "pwd" https://github.com/cyrille-leclerc/opentelemetry-plugin/commit/191cf48f7b3633b156a4518eef03ffe25fe93ad1

PR #951, opened 2021-02-04
chore: bump Go version to 1.15.7

kuisathaverat:
## What does this PR do?

Bump the Go version to 1.15.7


v1v (approved):

v1v:
Quick question, this particular version is the one provided by the `Beats` project?

kuisathaverat:
>Quick question, this particular version is the one provided by the Beats project? So, if they change we need to bump the version too?

We have to be alined on all projects, on those we cannot they have to use fixed versions with GO_VERSION env var or `.go-version` file

PR #952, opened 2021-02-04
feat: Archive only on fail

kuisathaverat:
## What does this PR do?

* new step isBuildFailure
* implement archiveOnlyOnFail for filebeat step
* add some missing annotations on the Axis step tests

## Why is it important?

Sometimes, we need to archive the Docker logs when there are errors, this new parameter allow that.

### DIFF
--- a/vars/isBuildFailure.groovy
+++ b/vars/isBuildFailure.groovy
@@ -1,0 +1,23 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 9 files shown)


vars/isBuildFailure.groovy
```diff
@@ -0,0 +1,23 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+  Return true if the build status is FAILURE or UNSTABLE
+*/
+def call() {
+  return currentBuild.result == 'FAILURE' || currentBuild.result == 'UNSTABLE'
```
v1v, line 22: NIT, though we might need to implement the function `resultIsWorseOrEqualTo` if it's not provided by the test library

```suggestion
  return currentBuild.resultIsWorseOrEqualTo('UNSTABLE')
```

v1v (approved):

vars/isBuildFailure.groovy
kuisathaverat: currentBuild is a variable in the test binding, this is stored in a Map, to allow call this method we have to introduce unnecessary complexity in the tests, so I reverted the change.

cachedout (approved):

PR #953, opened 2021-02-04
fix: resize the partition size to 100GB to use all the space

kuisathaverat:
## What does this PR do?

* add some commands to resize the size of the main partition.

## Why is it important?

We are using only 12GB of 100GB


mdelapenya (approved):

v1v (approved):
I'll work on enabling the Google Compute Plugin for our local instance:
- https://github.com/elastic/apm-pipeline-library/pull/356

Then we can benefits of ephemeral workers instead vagrant

PR #967, opened 2021-02-08
fix: enable sandbox for pipelines

kuisathaverat:
## What does this PR do?

* It enables sandbox on pipelines.

## Why is it important?

Without the sandbox flag enabled, jobs would need approval to be executed.

### DIFF
--- a/.ci/jobs/jobdsl-generator.yml
+++ b/.ci/jobs/jobdsl-generator.yml
@@ -4,6 +4,15 @@
     description: This job process the *.groovy files in .ci/jobDSL folder and create/update/delete jobs.
     view: APM-CI
     project-type: pipeline
+    properties:
+        - raw:
+            xml: |
+              <org.jenkinsci.plugins.authorizeproject.AuthorizeProjectProperty plugin="authorize-project@1.3.0">
+                <strategy class="org.jenkinsci.plugins.authorizeproject.strategy.SpecificUsersAuthorizationStrategy">
+                  <userid>local-admin</userid>
+                  <dontRestrictJobConfiguration>false</dontRestrictJobConfiguration>
+                </strategy>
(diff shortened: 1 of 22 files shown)


.ci/jobs/jobdsl-generator.yml
```diff
@@ -4,6 +4,15 @@
     description: This job process the *.groovy files in .ci/jobDSL folder and create/update/delete jobs.
     view: APM-CI
     project-type: pipeline
+    properties:
```
v1v, line 7: I'm a bit confused about this, what's the reason to run as admin?

v1v (approved):

cachedout (approved):

.ci/jobs/jobdsl-generator.yml
kuisathaverat: Since jobDSL 1.60, if you have security enabled, jobDSL will run with the security-script-plugin, this implies scripts have to be approved if you run outside a sandbox. 

```
[2021-02-08T16:15:37.632Z] Processing DSL script src/github.com/elastic/apm-pipeline-library/.ci/jobDSL/jobs/customKibana.groovy

script not yet approved for use
```

Then the proper way is to run them in a sandbox, because the job creates jobs it needs permissions to configure, our users do not have this permission, so the build fails with the following error:

```
You must configure the DSL job to run as a specific user in order to use the Groovy sandbox.
```

So you have two options here to disable the script-security-plugin for jobDSL or run the job as a user with configuration permissions. Disable the script-security-plugin for jobDSL needs changes in the provisioning system to disable the setting at the start time, which I think is not the best option, to run as another user you only have to use the authorization plugin to run this job with that user.

All of this does not happen if the Jenkins instance is configured with JCasC because jobDSL scripts are run as SYSTEM.

PR #970, opened 2021-02-09
Fix broken documentation

cachedout:
## What does this PR do?

Fixes instructions which do not work due to a typo in the path.


local/configs/jenkins.yaml
```diff
@@ -107,6 +107,7 @@ unclassified:
     endpoint: "otel-collector-contrib:4317"
     useTls: false
 jobs:
+  - file: ".ci/jobs/apm-shared.yml"
```
v1v, line 110: is this intended? If so , you might need to use full path

cachedout: Grr. that snuck in from my local dev. I'll remove it.

kuisathaverat (approved):

mdelapenya (approved):

PR #977, opened 2021-02-11
Disable/Enable pre-commit skipped tests in the report

v1v:
## What does this PR do?

Configure whether to enable/disable the pre-commited skipped checks as reported in the JUnit.

## Why is it important?

@SergeyKleyman found that the sanity-checks with skipped linted checks were a bit misleading since they are aggregated in the same test report with some other UTs/ITs....

Then we found out that it's worthy to use a red/green signal for this particular pre-commit checks.

### DIFF
--- a/vars/preCommitToJunit.groovy
+++ b/vars/preCommitToJunit.groovy
@@ -23,6 +23,7 @@
 def call(Map args = [:]) {
   def input = args.containsKey('input') ? args.input : error('preCommitToJunit: input parameter is required.')
   def output = args.containsKey('output') ? args.output : error('preCommitToJunit: output parameter is required.')
+  def reportSkipped = args.get('reportSkipped', false)
 
   def content = readFile(file: input)
   def id, status, message = '', inprogress = false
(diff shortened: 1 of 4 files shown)


vars/preCommitToJunit.groovy
```diff
@@ -42,16 +43,16 @@ def call(Map args = [:]) {
...
-    data += toJunit(id, status, message)
+    data += toJunit(id, status, message, reportSkipped)
   }
   data += '</testsuite>'
 
   writeFile file: output, text: data, encoding: 'UTF-8'
 }
 
-def toJunit(String name, String status, String message) {
+def toJunit(String name, String status, String message, reportSkipped=false) {
```
mdelapenya, line 53: Is it redundant to set this to false, as the caller already sets the default value to false?

v1v: I wanted to play safe, in case someone uses this function, theoretically it's something internal

mdelapenya (approved):

kuisathaverat (approved):

PR #979, opened 2021-02-12
[JCasC] google cloud engine local setup

v1v:
## What does this PR do?

Enable google cloud engine setup with JCasC for the local jenkins instance

## Why is it important?

Move away from the vagrant boxes.

## Related issues

Reuses https://github.com/elastic/apm-pipeline-library/pull/356
Notifies elastic/observability-dev#582


kuisathaverat (approved):

docs/CLOUD.md
```diff
@@ -0,0 +1,33 @@
+# Cloud
+
+We are using the Google Compute Jenkins plugins.
+
+See https://github.com/jenkinsci/google-compute-engine-plugin/#configuration-as-code-support for further examples.
+
+## How to
```
cachedout, line 7: IMHO we need to add a section here explaining what exactly this is and why somebody would want to configure it. The purpose of setting up GCE locally is not clear from reading this documentation.

mdelapenya (approved):

docs/CLOUD.md
v1v: I added the why in the description of this PR, how do you foresee we can improve this?

PR #1008, opened 2021-02-24
gsutil for windows with python

v1v:
## What does this PR do?

Install the google cloud sdk with Python bundled for Windows

## Why is it important?

Some Windows versions got issues with the python2.exe

image

Though we are not the only ones suffering this issue -> https://stackoverflow.com/questions/64289902/gsutil-says-it-cannot-find-python

Python 2 is correctly configured:

image

## Test

And the bundled python has been validated in a W8 VM:

image

### DIFF
--- a/vars/gsutil.groovy
+++ b/vars/gsutil.groovy
@@ -73,6 +73,7 @@ def googleCloudSdkURL() {
   if (isUnix()) {
     return "${url}-linux-${arch}.tar.gz"
   } else {
+    // use the bundled python artifact to avoid issues with the existing python2 installation when running gsutil in some Windows versions
     return "${url}-windows-${arch}-bundled-python.zip"
   }
 }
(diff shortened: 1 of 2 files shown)


cachedout (approved):

vars/gsutil.groovy
```diff
@@ -69,9 +69,12 @@ def downloadWithCurl(tarball, url) {
...
 def googleCloudSdkURL() {
   def url = 'https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-319.0.0'
-  def os = isUnix() ? 'linux' : 'windows'
-  def ext = isUnix() ? 'tar.gz' : 'zip'
-  return "${url}-${os}-${is64() ? 'x86_64' : 'x86'}.${ext}"
+  def arch = is64() ? 'x86_64' : 'x86'
+  if (isUnix()) {
+    return "${url}-linux-${arch}.tar.gz"
+  } else {
+    return "${url}-windows-${arch}-bundled-python.zip"
```
mdelapenya, line 76: Maybe a comment here explaining the need of the bundled python installer could help in future lookups

mdelapenya (approved):

vars/gsutil.groovy
v1v:
```suggestion
    // use the bundled python artifact to avoid issues with the existing python2 installation when running gsutil in some Windows versions
    return "${url}-windows-${arch}-bundled-python.zip"
```

PR #1015, opened 2021-03-05
Remove dependency with https://plugins.jenkins.io/junit-attachments/

v1v:
## What does this PR do?

Remove https://plugins.jenkins.io/junit-attachments/ dependency

## Why is it important?

https://plugins.jenkins.io/junit-attachments/ is in adoption, and let's remove a plugin that probably we won't need

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -175,12 +175,7 @@ pipeline {
       }
       post {
         always {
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):
in this pipeline, we do not need the plugin. This is only required where we generated screenshots or similar outputs

PR #1016, opened 2021-03-05
Update to new schema for flaky tests

cachedout:
## What does this PR do?

This is the first part of a request to update the APM Pipeline library to support the new flaky test schema.

It is not ready for merge. Still needs tests and to have a few items double-checked, such as whether or not the test names are aligned with the names of the actual test failures flowing into this function.

## Why is it important?

Needed to restore flaky test functionality to PRs.


src/co/elastic/NotificationManager.groovy
```diff
@@ -60,32 +58,40 @@ def analyzeFlakey(Map args = [:]) {
...
+
       def foundFlakyList = testFlaky?.size() > 0 ? testFailures.intersect(testFlaky) : []
       genuineTestFailures = testFailures.minus(foundFlakyList)
-      log(level: 'DEBUG', text: "analyzeFlakey: Flaky tests raw: ${flakeyTestsRaw}")
-      log(level: 'DEBUG', text: "analyzeFlakey: Flaky matched tests: ${foundFlakyList.join('\n')}")
+      log(level: 'INFO', text: "analyzeFlakey: Flaky tests raw: ${flakeyTestsRaw}")
+      log(level: 'INFO', text: "analyzeFlakey: Flaky matched tests: ${foundFlakyList.join('\n')}")
+
+      // FIXME this exits early to avoid creating actual issues during development
+      return
```
cachedout, line 94: **DANGER**: For development only! Remove before merge!

src/co/elastic/NotificationManager.groovy
```diff
@@ -367,19 +373,57 @@ def generateBuildReport(Map args = [:]) {
...
+  "query": {
+    "bool": {
+      "must": [],
+      "filter": [
+        {
+          "bool": {
+            "should": [
+              {
+                "match_phrase": {
+                  "job.fullName": "Beats/beats/master"
```
v1v, line 407: Maybe the env variable JOB_NAME could help here:

>JOB_NAME
Name of the project of this build, such as "foo" or "foo/bar".


For instance:

```
$ curl -s 'https://beats-ci.elastic.co/job/Beats/job/beats/job/master/lastSuccessfulBuild/artifact/env-info.json' | jq .JOB_NAME
"Beats/beats/master"
```

Although IIRC, the flaky test analyser only worked for the master branch. so if working with a PR I'd say it should then be forced to the master branch

src/co/elastic/NotificationManager.groovy
```diff
@@ -367,19 +373,57 @@ def generateBuildReport(Map args = [:]) {
...
-                  { "timestamp" : "desc" },
-                  { "test_score" : "desc" }
-                ],
-                "query" : {
-                  "range" : {
-                    "test_score" : {
-                      "gt" : ${flakyThreshold}
-                    }
-                  }
+def queryFilter(timeout) {
```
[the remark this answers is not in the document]
cachedout: Indeed! I don't think we'll need that going forward so I've removed it. Thanks for catching this.

src/co/elastic/NotificationManager.groovy
cachedout: Ah! Very nice! Thanks.

src/co/elastic/NotificationManager.groovy
cachedout: We faced a similar problem in the past which is how we ended up with [flakyReportIdx](https://github.com/search?p=1&q=org%3Aelastic+flakyReportIdx&type=Code), which allowed individual pipelines to select the index that they wanted the Notification Manager class to do their Elasticsearch lookup against.

It's a bit disruptive, but what I think we should do here is to immediately deprecate `flakyReportIdx` as an option and raise a warning if it is passed in. Then we should add a new option called `jobName`. Consumers of the step would typically pass in the `JOB_NAME` variable as you suggest, but in certain cases (such as Beats), [this could be modified](https://github.com/elastic/beats/blob/024dd5c98021368bda3c7b8b0a91f69c9b14d5aa/Jenkinsfile#L169-L176). 

@v1v WDYT?

cachedout: I have updated this to do the following:

1. Log a warning and exit early if just `flakyReportIdx` is passed in.
2. Error if both `flakyReportIdx` and `jobName` are used.
3. Error if neither `flakyReportIdx` nor `jobName` are used.

vars/notifyBuildResult.txt
```diff
@@ -35,8 +35,6 @@ emails on Failed builds that are not pull request.
 * slackCredentials: What slack credentials to be used. Default value uses `jenkins-slack-integration-token`.
 * slackNotify: Whether to send or not the slack notifications, by default it sends notifications on Failed builds that are not pull request.
 * analyzeFlakey: Whether or not to add a comment in the PR with tests which have been detected as flakey. Default: `false`.
```
v1v, line 37: Missing the `jobName`

vars/notifyBuildResult.groovy
```diff
@@ -75,8 +73,6 @@ def call(Map args = [:]) {
         data['statsUrl'] = statsURL
         data['es'] = es
         data['es_secret'] = secret
```
v1v, line 75: missing the `jobName` -> https://github.com/elastic/apm-pipeline-library/pull/1016/files#diff-f9698af828252702ef720a3af41ec8d7469294e5149a71ac4edf48c283a14834R47

src/co/elastic/NotificationManager.groovy
```diff
@@ -35,8 +35,6 @@ This method generates flakey test data from Jenkins test results
  * @param secret Vault path to secrets which hold authentication information for Elasticsearch
  * @param jobInfo JobInfo data collected from job-info.json
  * @param testsErrors list of test failed, see src/test/resources/tests-errors.json
```
v1v, line 37: Missing the `jobName`

src/co/elastic/NotificationManager.groovy
```diff
@@ -60,32 +57,44 @@ def analyzeFlakey(Map args = [:]) {
     def flakyTestsWithIssues = [:]
     def genuineTestFailures = []
 
-    if (!flakyReportIdx?.trim()) {
-      error 'analyzeFlakey: did not receive flakyReportIdx data'
+    if (args.containsKey('flakyReportIdx') && args.containsKey('jobName')) {
+      error('Please pass either `jobName` or `flakyReportIdx` but not both')
```
v1v, line 61: I'm a bit skeptical with this particular validation, since this is a breaking change. 

We are forced to change the consumers anyway, what do you think if we remove entirely the flakyReportIdx argument and make the `jobName` as mandatory?

cachedout: The idea here was to avoid breaking any pipelines during the transition period. If you think that won't be a problem, we can definitely simplify this.

vars/notifyBuildResult.txt
cachedout: Whoops. Forgot to push that commit. Thanks!

src/co/elastic/NotificationManager.groovy
v1v: I missed your previous comment, sorry about that.

https://github.com/elastic/apm-pipeline-library/blob/ce21387c1d2157a5528321c3719f45bc4e814485/vars/notifyBuildResult.groovy#L193-L201 will not fail but warn the stage only, so I think we are safe even if we force the `jobName`.

what do you think?

cachedout: I think you make a great point here and I prefer your approach. I'll back out my changes and we'll rely on the handling in `notifyBuildResult` to do the right thing. Thanks for raising this issue!

v1v:
https://github.com/elastic/beats/pull/24464 might be the changes required in the consumers of this step

src/co/elastic/NotificationManager.groovy
cachedout: I have removed this check per https://github.com/elastic/apm-pipeline-library/pull/1016#discussion_r590522960

src/co/elastic/NotificationManager.groovy
cachedout: I've backed my changes out. Thanks for the discussion here. It was really helpful!

src/co/elastic/NotificationManager.groovy
```diff
@@ -35,57 +35,55 @@ This method generates flakey test data from Jenkins test results
...
-    if (!flakyReportIdx?.trim()) {
-      error 'analyzeFlakey: did not receive flakyReportIdx data'
-    }
-
-    // Only if there are test failures to analyse
-    if(testsErrors.size() > 0) {
+    // 1. Only if there are test failures to analyse
+    // 2. Only continue if we have a jobName passed in. This does not raise an error to preserve
+    // backward compatability.
+    if(testsErrors.size() > 0 && jobName != null) {
```
v1v, line 62: L47 will throw an error, so I guess the jobName will never be null but maybe empty (https://github.com/elastic/apm-pipeline-library/pull/1016/files#diff-e79413d8e910494d1fa40002bd01bb66c3830ab0646a68e93e15c4d43a7a22e7R69)?

```suggestion
    if(testsErrors.size() > 0 && jobName?.trim()) {
```

src/co/elastic/NotificationManager.groovy
```diff
@@ -35,57 +35,55 @@ This method generates flakey test data from Jenkins test results
...
+      //
+      //def testFlakyFreq = [:]
+      //flakeyTestsParsed?.aggregations?.test_name?.buckets?.each { it -> testFlakyFreq[it['key']] = it['doc_count'] } 
+
       def foundFlakyList = testFlaky?.size() > 0 ? testFailures.intersect(testFlaky) : []
       genuineTestFailures = testFailures.minus(foundFlakyList)
-      log(level: 'DEBUG', text: "analyzeFlakey: Flaky tests raw: ${flakeyTestsRaw}")
-      log(level: 'DEBUG', text: "analyzeFlakey: Flaky matched tests: ${foundFlakyList.join('\n')}")
+      log(level: 'INFO', text: "analyzeFlakey: Flaky tests raw: ${flakeyTestsRaw}")
+      log(level: 'INFO', text: "analyzeFlakey: Flaky matched tests: ${foundFlakyList.join('\n')}")
```
v1v, line 86: Maybe too verbose by default?

```suggestion
      log(level: 'DEBUG', text: "analyzeFlakey: Flaky tests raw: ${flakeyTestsRaw}")
      log(level: 'DEBUG', text: "analyzeFlakey: Flaky matched tests: ${foundFlakyList.join('\n')}")
```

src/co/elastic/NotificationManager.groovy
```diff
@@ -367,19 +365,57 @@ def generateBuildReport(Map args = [:]) {
...
-                "query" : {
-                  "range" : {
-                    "test_score" : {
-                      "gt" : ${flakyThreshold}
-                    }
-                  }
+def queryFilter(jobName) {
+  return """
+{
+  "aggs": {
```
v1v, line 371: Do we need the default values such as `should`, `script_fields` ...?

cachedout: I massaged the query down a bit to just what I think should be needed.

v1v (approved):

PR #1025, opened 2021-03-08
add noTags - fixes #1024

### LINKED ISSUE #1024: gitCheckout add noTags
gitCheckout forces `noTags` to false. It would be great if we could surface that option.

happy to write a patch

tarekziade:
## What does this PR do?

Adds the `noTags`  option in `gitCheclout`

## Why is it important?

We are using the apm lib in all our jobs, and we have a custom checkout function that could go away if `gitCheckout` let us chec out wih no tags

## Related issues
Closes #1024


v1v (approved):
Thanks for the contributions 💯

kuisathaverat (approved):

PR #1031, opened 2021-03-11
chore: support filtering the upstream cause in the isUpstreamTrigger step

mdelapenya:
## What does this PR do?
It updates the `isUpstreamTrigger` step to identify whether the upstream trigger was caused by a PR or not, passing a filter in a backwards compatible mode.

As @v1v suggested during the review, consumers will use the step exactly the same as before, but now supporting a `filter` string parameter:

```groovy
whenTrue(isUpstreamTrigger()) {
   //
}

whenTrue(isUpstreamTrigger(filter: 'PR-')) {
  //
}
```

## Why is it important?
In certain cases, we want the downstream job to act in a different manner for PRs than for branches, like selecting a subset of the downstream project's test suite.

## Related issues
Relates elastic/e2e-testing#873


vars/isUpstreamPRTrigger.groovy
```diff
@@ -0,0 +1,31 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+  Check it the build was triggered by an upstream job from a pull request.
+
+  def upstreamPRTrigger = isUpstreamPRTrigger()
+*/
+def call(){
+  def buildCause = currentBuild.getBuildCauses()?.find{ it._class == 'hudson.model.Cause$UpstreamCause'}
```
mdelapenya, line 24: I was tempted to call `isUpstreamTrigger` first, but then thought about the cost of iterating through all the build causes twice.

v1v (approved):

vars/isUpstreamPRTrigger.groovy
v1v: I was thinking to change the signature of `isUpstreamTrigger` and support the filter something like

```
// isUpstreamTrigger
def call(Map args=[:]) {
  def filter = args.get('filter', 'all') 
  def buildCause = currentBuild.getBuildCauses()?.find{ it._class == 'hudson.model.Cause$UpstreamCause'}
  if (buildCause?.upstreamProject?.trim()) {
     // evaluate filter for PR
     if (filter!= 'all' ) {
       return buildCause?.upstreamProject?.toUpperCase()?.contains(filter))
     }
     return true
  }
  return false
}

```


the consumers

```

whenTrue(isUpstreamTrigger()) {
   //
}

whenTrue(isUpstreamTrigger(filter: 'PR-')) {
  //
}
```

cachedout (approved):

vars/isUpstreamPRTrigger.groovy
mdelapenya: Do you mind if I add a commit rolling back this one, and adding that logic to the `isUpstreamTrigger` step?

vars/isUpstreamTrigger.groovy
```diff
@@ -16,14 +16,22 @@
...
+def call(Map args=[:]){
+  def filter = args.get('filter', 'all')
+
   def buildCause = currentBuild.getBuildCauses()?.find{ it._class == 'hudson.model.Cause$UpstreamCause'}
   if (buildCause?.upstreamProject?.trim()) {
-    log(level: 'DEBUG', text: "isUpstreamTrigger: ${buildCause?.upstreamProject?.toString()}")
+    log(level: 'DEBUG', text: "isUpstreamTrigger: ${buildCause?.upstreamProject?.toString()}, filter: '${filter}'")
+    // evaluate filter
+    if (filter != 'all' ) {
+      return buildCause?.upstreamProject?.toUpperCase()?.contains(filter)
```
kuisathaverat, line 32:
```suggestion
      return buildCause?.upstreamProject?.toUpperCase()?.contains(filter.toUpperCase())
```

kuisathaverat (approved):

vars/isUpstreamTrigger.groovy
mdelapenya: Good catch!

vars/isUpstreamTrigger.groovy
```diff
@@ -16,14 +16,22 @@
...
+def call(Map args=[:]){
+  def filter = args.get('filter', 'all')
+
   def buildCause = currentBuild.getBuildCauses()?.find{ it._class == 'hudson.model.Cause$UpstreamCause'}
   if (buildCause?.upstreamProject?.trim()) {
-    log(level: 'DEBUG', text: "isUpstreamTrigger: ${buildCause?.upstreamProject?.toString()}")
+    log(level: 'DEBUG', text: "isUpstreamTrigger: ${buildCause?.upstreamProject?.toString()}, filter: '${filter}'")
+    // evaluate filter
+    if (filter != 'all' ) {
+      return buildCause?.upstreamProject?.toUpperCase()?.contains(filter.toUpperCase())
```
v1v, line 32: capital case? or ignore-case?

v1v: I said nothing!

vars/isUpstreamTrigger.txt
```diff
@@ -1,5 +1,6 @@
-Check if the build was triggered by an upstream job.
+Check if the build was triggered by an upstream job, being it possible to add some filters.
 
 ```
 def upstreamTrigger = isUpstreamTrigger()
+def upstreamTrigger = isUpstreamTrigger(filter: 'PR-')
```
v1v, line 5: Missing the new argument and the description

v1v (commented):
+1 after the docs change

vars/README.md
```diff
@@ -1556,21 +1556,22 @@ Whether the build is based on a Tag Request or no
...
-Check it the build was triggered by a timer (scheduled job).
+Check if the build was triggered by a timer (scheduled job).
 
 ```
 def timmerTrigger = isTimerTrigger()
 ```
 
 ## isUpstreamTrigger
-Check if the build was triggered by an upstream job.
+Check if the build was triggered by an upstream job, being it possible to add some filters.
```
mdelapenya, line 1566:
```suggestion
Check if the build was triggered by an upstream job, being it possible to add a filter for the upstream cause.
```

vars/isUpstreamTrigger.groovy
```diff
@@ -16,14 +16,22 @@
 // under the License.
 
 /**
-  Check it the build was triggered by an upstream job.
+  Check if the build was triggered by an upstream job, being possible to add some filters.
```
mdelapenya, line 19:
```suggestion
  Check if the build was triggered by an upstream job, being possible to add a filter for the upstream cause.
```

vars/isUpstreamTrigger.txt
mdelapenya: I added it and also improved the description, thanks!

PR #1049, opened 2021-03-23
chore: refactor filebeat step

kuisathaverat:
## What does this PR do?

* Minor refactor of the output variable use
* Avoid hiding errors when the container is not launched.
* * Bump the container version

## Why is it important?

When an error happens during the execution of the Docker container, it always shows an error reading the config file but there is one error before that.


v1v (approved):

mdelapenya (commented):
+1, although there is a test error

cachedout (approved):

PR #1061, opened 2021-03-29
feat: stage status cache

kuisathaverat:
## What does this PR do?

It adds a new step to save/check the last execution status of a stage-based in its name and commit.

## Why is it important?

This allows skipping stages for the same commit in case of retries.

### DIFF
--- a/vars/stageStatusCache.groovy
+++ b/vars/stageStatusCache.groovy
@@ -28,9 +28,9 @@ def call(Map args, Closure body){
     args['credentialsId'] = 'beats-ci-gcs-plugin-file-credentials'
   }
   if(!args.containsKey('sha')){
-    args['sha'] = "${GIT_BASE_COMMIT}"
+    args['sha'] = "${env.GIT_BASE_COMMIT}"
   }
-  if(readStageStatus(args) == false || isUserTrigger() || env.BUILD_ID == "1" || args.get('runAlways', false)){
+  if(env.BUILD_ID == "1" || isUserTrigger() || args.get('runAlways', false) || readStageStatus(args) == false){
     body()
     saveStageStatus(args)
   } else {
@@ -59,8 +59,6 @@ def readStageStatus(Map args){
     cmd(label: 'Download Stage Status',
       script: "curl -sSf -O https://storage.googleapis.com/${args.bucket}/ci/cache/${statusFileName}",
       returnStatus: true)
-  } catch(e) {
-    log(level: 'WARN', text: "There is no cache file for the current stage.")
   } finally {
     return fileExists("${statusFileName}")
   }
@@ -70,5 +68,5 @@ def readStageStatus(Map args){
   generate an unique ID for the stage and commit.
 */
 def stageStatusId(Map args){
-  return base64encode(text: "${args.id}${args.sha}", encoding: "UTF-8")
+  return base64encode(text: "${args.id}${args.sha}", encoding: "UTF-8", padding: false)
 }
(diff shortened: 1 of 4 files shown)


cachedout (approved):

vars/stageStatusCache.groovy
```diff
@@ -0,0 +1,74 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+  Execute the body if there is not stage status file for the stage and the commit we are building.
+  User triggered builds will execute all stages always.
+  If the stage success the status is save in a file.
+*/
+def stageStatusCache(Map args, Closure body){
```
v1v, line 23:
```suggestion
def call(Map args, Closure body){
```

vars/stageStatusCache.groovy
```diff
@@ -0,0 +1,74 @@
...
+  if(!args.containsKey('bucket')){
+    args['bucket'] = 'beats-ci-temp'
+  }
+  if(!args.containsKey('credentialsId')){
+    args['credentialsId'] = 'beats-ci-gcs-plugin-file-credentials'
+  }
+  if(!args.containsKey('sha')){
+    args['sha'] = "${GIT_BASE_COMMIT}"
+  }
+  if(readStageStatus(args) == false || isUserTrigger() || env.BUILD_ID == "1" || env.RUN_ALL == "1"){
```
v1v, line 33: `RUN_ALL` is quite specific, do you think we can use an argument instead? Something like

```suggestion
  if(readStageStatus(args) == false || isUserTrigger() || env.BUILD_ID == "1" || args.get('runAlways', false)){
```

vars/stageStatusCache.groovy
```diff
@@ -0,0 +1,74 @@
...
+  Read the status file of the stage if it exists.
+*/
+def readStageStatus(Map args){
+  def statusFileName = stageStatusId(args)
+  try {
+    cmd(label: 'Download Stage Status',
+      script: "curl -sSf -O https://storage.googleapis.com/${args.bucket}/ci/cache/${statusFileName}",
+      returnStatus: true)
+  } catch(e) {
+    log(level: 'WARN', text: "There is no cache file for the current stage.")
```
v1v, line 63: I guess, this should never happen as long as the `returnStatus: true` is set

vars/stageStatusCache.groovy
v1v, line 23: Missing the list of supported arguments:

- bucket
- credentialsId
- sha
- id   < ---- this should be mandatory

vars/stageStatusCache.txt
```diff
@@ -0,0 +1,33 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+* *bucket:* bucket name.
+* *credentialsId:* credentials file, with the GCP credentials JSON file.
+* *sha:* Commit SHA used for the stage ID.
```
v1v, line 33:
```suggestion
* *id:* Unique stage name. Mandatory
* *bucket:* bucket name. Default 'beats-ci-temp'
* *credentialsId:* credentials file, with the GCP credentials JSON file. Default  'beats-ci-gcs-plugin-file-credentials'
* *sha:* Commit SHA used for the stage ID. Default: env.GIT_BASE_COMMIT
```

vars/stageStatusCache.groovy
v1v: https://github.com/elastic/apm-pipeline-library/pull/1061#discussion_r603894032

vars/stageStatusCache.groovy
kuisathaverat: yep

v1v (approved):

PR #1074, opened 2021-04-14
Fix build status and adjust Vault

cachedout:
## What does this PR do?

This contains a variety of fixes which related to the Jenkins release job for the apm-java-agent project.

### Build Status

The previous step used the standard library to make HTTP requests. However, it was silently failing and returning incorrect results. We change to use our existing HTTP client implementation which works more reliably and returns the correct result.

### Make vault credentials configurable

The `internal-ci` machine is shared between multiple teams and our credentials are namespaced accordingly. Therefore, I modified the Vault implementation to make it possible to pass in specific IDs to use when authenticating to Vault.

## Why is it important?

Fixes issues related to the broken release job for the APM Java agent.

## Related issues
https://github.com/elastic/observability-robots/issues/463

### DIFF
--- a/vars/getVaultSecret.groovy
+++ b/vars/getVaultSecret.groovy
@@ -47,7 +47,7 @@ def call(secret) {
 def readSecret(secret, role_id, secret_id) {
   def props = null
   log(level: 'INFO', text: 'getVaultSecret: Getting secrets')
-  readSecretWrapper('vault-role-id', 'vault-secret-id') {
+  readSecretWrapper(role_id, secret_id) {
     // When running in the CI with multiple parallel stages
     // the access could be considered as a DDOS attack. Let's sleep a bit if it fails.
     retryWithSleep(retries: 3, seconds: 5, backoff: true) {
--- a/vars/getVaultSecret.groovy
+++ b/vars/getVaultSecret.groovy
@@ -47,7 +47,7 @@ def call(secret) {
 def readSecret(secret, role_id, secret_id) {
   def props = null
   log(level: 'INFO', text: 'getVaultSecret: Getting secrets')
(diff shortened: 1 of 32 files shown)


vars/getVaultSecret.groovy
```diff
@@ -58,11 +60,11 @@ def readSecret(secret) {
   return props
 }
 
-def readSecretWrapper(body) {
+def readSecretWrapper(role_id, secret_id, body) {
```
kuisathaverat, line 63: this change breaks a few steps and pipelines https://github.com/search?q=org%3Aelastic+readSecretWrapper&type=code

cachedout: Hrm. OK. I guess we'll have to go ahead and create a diverging code path for the needed params, then. I'll do that and restore the original method signature.

kuisathaverat (approved):

vars/getVaultSecret.groovy
```diff
@@ -58,6 +60,20 @@ def readSecret(secret) {
   return props
 }
 
+def readSecretWrapperWithParams(role_id, secret_id, body) {
```
v1v, line 63: If possible I'd say to use the `Maps args=[:]` this should help us to add parameters in the future without affecting the signature and therefore the consumers

cachedout: I update the signature to `(args, body)`.

vars/getVaultSecret.groovy
cachedout: Done.

v1v (approved):
I did a quick search for https://github.com/search?l=Groovy&q=org%3Aelastic+readSecret&type=Code and I couldn not find any other consumers
+1

PR #1077, opened 2021-04-15
feat: step to check if a docker image exists

mdelapenya:
## What does this PR do?
It adds a new step that will check if a given Docker image exists, checking the Docker host first: if that fails because the image does not exist in the current Docker host, then it will try to look up the manifest in the Docker registry. If it fails again then it means the image does not exist, it will log an error and will return false.

## Why is it important?
We want to skip building Kibana again and again, if the image was already built. Specially important when we want to trigger an action on the same commit (i.e. triggering the e2e tests)


vars/checkDockerImage.groovy
```diff
@@ -0,0 +1,39 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+  Checks if the given Docker image exists.
+
+  checkDockerImage(image: 'hello-world:latest')
+*/
+def call(Map args = [:]) {
```
v1v, line 23: should return a boolean?

vars/checkDockerImage.groovy
```diff
@@ -0,0 +1,39 @@
...
+  def redirectStdout = isUnix() ? '2>/dev/null' : '2>NUL'
+  if (cmd(returnStatus: true, script: "docker images -q ${image}  ${redirectStdout}") == 0) {
+    log(level: 'DEBUG', text: "${image} exists in the Docker host")
+    return
+  }
+
+  log(level: 'DEBUG', text: "${image} does not exist: pulling")
+  if (cmd(returnStatus: true, script: "docker pull ${image}") == 0) {
+    return
+  }
```
v1v, line 35: What do you think to add a default new argument to try with pulling?

```
checkDockerImage(image: 'hello-world:latest', pull: true)
```

It can be enabled by default. Then this step can verify the cached docker images or also an image from an external docker registry.

Besides, do you wanna pull or just query if it's available?

mdelapenya: The bool return is in a upcoming commit!

mdelapenya: I just want to check. Maybe there is a simpler/faster manner

vars/checkDockerImage.groovy
```diff
@@ -0,0 +1,41 @@
...
+  def image = args.containsKey('image') ? args.image : error('checkDockerImage: image parameter is required')
+
+  def redirectStdout = isUnix() ? '2>/dev/null' : '2>NUL'
+  if (cmd(returnStatus: true, script: "docker images -q ${image}  ${redirectStdout}") == 0) {
+    log(level: 'DEBUG', text: "${image} exists in the Docker host")
+    return true
+  }
+
+  log(level: 'DEBUG', text: "${image} does not exist: pulling")
+  if (cmd(returnStatus: true, script: "docker pull ${image}") == 0) {
```
cachedout, line 35: This is really surprising behavior, tbh. I would not have expected a step which says it will check the status of something to have the side-effect of changing the state of the system.

IMHO this should either be delegated to a step dedicated to pulling an image or it should be put behind a flag which is disabled by default. (Something like `pullIfNotFound: false`. WDYT?

mdelapenya: In my past experience, check methods were related to check that a state was satisfied, performing the operations to satisfy it if/when needed, that's why I named this step after `check`. In fact I was tempted to create to steps, something like `isDockerImagePulled` and `pullDockerImage` (name them), calling both sequentially. But OTOH I like the name you propose `pullIfNotFound`, as it's more straightforward and clearer to understand its purpose.

mdelapenya: I meant the name of the attribute (`pullIfNotFound`), although the name of the step should be clearer too

vars/checkDockerImage.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+  whenTrue(checkDockerImage(image: 'hello-world:latest')) {
+    ...
+  }
+*/
+def call(Map args = [:]) {
+  def image = args.containsKey('image') ? args.image : error('checkDockerImage: image parameter is required')
+
+  def redirectStdout = isUnix() ? '2>/dev/null' : '2>NUL'
+  if (cmd(returnStatus: true, script: "docker images -q ${image}  ${redirectStdout}") == 0) {
```
v1v, line 29: What happen if docker is not installed? For instance, in CI `macOS` workers is required to prepare the docker environment context, otherwise docker won't be ready, IIRC.

You could potentially use `isInstalled(tool: 'docker', flag: '--version')` for that particular validation

mdelapenya: Cool, I think it's important to check the tools. Thanks for the suggestion, will submit a commit with it ASAP

mdelapenya: Added in dacf11682ab6a446ee5e89a9d860fab4c82861ea

vars/checkDockerImage.groovy
mdelapenya: @cachedout I added the new parameter, as you suggested. Still looking to a better name for the step, which I'd like to open the discussion.

I like checkXXX because it will try to satisfy the XXX if possible, in contrast to the getter methods (isXXX, getXXX) which simply returns the boolean value after verifying for the current state. In this particular case we have two steps to verify if a Docker image exists:

1. does the image live in the docker host?
2. does the image live in the docker registry?

Thinking out loud, we do not need to effectively pull the image, but simply return YES/NO for 2). I've an idea: using

```shell
docker manifest inspect docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT > /dev/null 
```

mdelapenya: With the `docker manifest` approach, I think the name of the step can be changed to `existsDockerImage`. Therefore, we won't need the `pullIfNotFound` attribute.

cachedout: > Still looking to a better name for the step, which I'd like to open the discussion.

`ensureDockerImageExists` ?

mdelapenya: now that we are not internally pulling, simply querying the docker manifest API, I think `dockerImageExists` is more accurate, as we are not changing the docker host state with a docker-pull command. If you agree, I'm going to rename the step to that name

cachedout: > If you agree, I'm going to rename the step to that name

Yup, makes sense. Thanks!

docs/DEVELOPMENT.md
```diff
@@ -77,6 +77,10 @@ execute the following from the directory at the root of this project:
 
 `./mvnw test -Dtest=DummyStepTests`
 
+To run one single test, execute the following from the directory at the root of
+this project, separating test name from test file with an `#`:
+
+`./mvnw test -Dtest=DummyStepTests#testName`
```
kuisathaverat, line 83:
```suggestion
`./mvnw test -P debug -Dtest=DummyStepTests#testName`
```

mdelapenya: We could add a separate section for running tests in debug mode. I simply extended the previous section. I can do it in a new commit.

mdelapenya: Well, I 've just read it is explained below 😊 (https://github.com/elastic/apm-pipeline-library/pull/1077/files#diff-d9657425e036fb0ad2912c2c24ea793648eb86de5b40aa0844276dd08ed01d87R85)

vars/dockerImageExists.groovy
```diff
@@ -0,0 +1,47 @@
...
+*/
+def call(Map args = [:]) {
+  if (!isInstalled(tool: 'docker', flag: '--version')) {
+    error 'Docker is not installed'
+  }
+
+  def image = args.containsKey('image') ? args.image : error('dockerImageExists: image parameter is required')
+
+  def redirectStdout = isUnix() ? '2>/dev/null' : '2>NUL'
+  if (cmd(returnStatus: true, script: "docker images -q ${image} ${redirectStdout}") == 0) {
```
kuisathaverat, line 33: The return code is always 0

```
docker images busybox:foo && echo "It does not work"
```

kuisathaverat: these commands works

```
docker inspect busybox:foo || echo "the image is not here"
```

this will return the Ids if exist locally
```
docker inspect -f "{{.Id}}" busybox:latest
```

mdelapenya: Ah good catch! Will update the command, thanks!

kuisathaverat (approved):

vars/dockerImageExists.groovy
```diff
@@ -0,0 +1,46 @@
...
+  whenTrue(dockerImageExists(image: 'hello-world:latest')) {
+    ...
+  }
+*/
+def call(Map args = [:]) {
+  if (!isInstalled(tool: 'docker', flag: '--version')) {
+    error 'Docker is not installed'
+  }
+
+  def image = args.containsKey('image') ? args.image : error('dockerImageExists: image parameter is required')
```
v1v, line 30: I'd say to move this parameter validation earlier, and add a UT for that:

```
  @Test
  void test_missing_image() throws Exception {
    testMissingArgument('image') {
      script.call()
    }
  }
```

vars/dockerImageExists.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+  if (cmd(returnStatus: true, script: "docker inspect -f \"{{.Id}}\" ${image}") == 0) {
+    log(level: 'DEBUG', text: "${image} exists in the Docker host")
+    return true
+  }
+
+  def redirectStdout = isUnix() ? '>/dev/null' : '>NUL'
```
v1v, line 37: I think you don't need this redirection if you use `cmd(returnStdout: true, ...)`

vars/dockerImageExists.groovy
v1v: Fail fast if wrongly configured instead running the `docker is installed` validation

v1v (approved):
Minor comments that are not blockers

mdelapenya:
> Minor comments that are not blockers

I addressed them in last 3 commits, thanks!

PR #1078, opened 2021-04-19
Enable to customise GitHub PR comments

v1v:
## What does this PR do?

1) Enable to use a customer GitHub PR comment template either with a file to be read or a text.
2) Provide a new argument with the list of step errors filtered by some generic ones.
3) Less verbose details in the flaky report.

## Why is it important?

This will help the consumers to specify what template they would like to use instead the default one. The template definition should be hosted in the consumes repo.

To enable the text template is required to use the argument `useContentTemplate`, this is helpful since the `notifyBuildResult` step runs in a clean worker then we don't need to `unstash` the workspace.

### 💚 

image

### Aborted

image

### Test failures

image


src/co/elastic/NotificationManager.groovy
```diff
@@ -317,6 +323,8 @@ def notifySlack(Map args = [:]) {
 /**
  * This method generates the build report, archive it and returns the build report
  * @param archiveFile whether to create and archive the file.
+ * @param template what build file template to be used.
```
v1v, line 326:
```suggestion
 * @param buildCommentTemplate what build file template to be used.
```

vars/notifyBuildResult.groovy
```diff
@@ -80,6 +80,15 @@ def call(Map args = [:]) {
         data['enabled'] = slackNotify
         data['jobName'] = jobName
 
+        // Allow to use either a file or a content of the file
+        // rationale: this stage does not checkout the source code, fileContentTemplate allows
+        // to use the content instead the file path.
+        data['buildCommentTemplate'] = args.get('buildCommentTemplate', 'github-comment-markdown.template')
+        if (args.get('useContentTemplate', false) && args.containsKey('buildCommentTemplate')) {
+          data['buildCommentTemplate'] = buildCommentTemplate
```
v1v, line 88:
```suggestion
```

PR #1093, opened 2021-04-27
Remove duplicated entries in artifacts-api-latest-versions.sh

v1v:
## What does this PR do?

`7.x` and `7.<minor>` had the same data structure and therefore the output had two keys with the same value.
Fix linting defects

## Why is it important?

Avoid duplicated entries when running the bump automation:

image

image


resources/scripts/artifacts-api-latest-versions.sh
```diff
@@ -48,8 +49,8 @@ for version in ${QUERY_OUTPUT}; do
...
         echo "${LATEST_OUTPUT}"
-    } >> "${OUTPUT}"
+    } >> "${TEMP_FILE}"
 done
-echo "}" >> "${OUTPUT}"
+echo "}" >> "${TEMP_FILE}"
 
-cat "${OUTPUT}"
\ No newline at end of file
+cat "${TEMP_FILE}" | jq | tee ${OUTPUT}
```
v1v, line 56:
```suggestion
cat "${TEMP_FILE}" | jq | tee ${OUTPUT}

```

mdelapenya (approved):

PR #1094, opened 2021-04-27
Fix jq format for linux

v1v:
## What does this PR do?

In linux the script does not work

image

image

While it worked fine in MacOS.

I tested in linux and mac now:

image

image


v1v:
I'll merge this PR otherwise the current automation will produce wrong data

PR #1102, opened 2021-05-03
fix: add a JAVA_OPTS env var to fix the container start

kuisathaverat:
## What does this PR do?

* It add a JAVA_OPT with the classpath to allow to use the latest Infra image.

## Why is it important?

The new image uses ECS and APM and those jar files are in a folder outside the classpath so you need to pass the classpath.


cachedout (approved):
Nice find!

mdelapenya (approved):

PR #1103, opened 2021-05-03
Revert "[opbeans-php] Add docker build and cache (#888)"

cachedout:
This reverts commit 2bc37595ee87c3d9dc4f6993788f048f6ec09f10.

## What does this PR do?
This is a full revert of https://github.com/elastic/apm-pipeline-library/pull/888

I believe that there should not actually be a build step for the Opbeans PHP application. I'll explain my reasoning below.

For the past several builds (and likely since [this PR was merged on January 5](https://github.com/elastic/apm-pipeline-library/pull/888)), the [builds for the Opbeans applications have been failing in the APM CI](https://apm-ci.elastic.co/job/apm-shared/job/apm-docker-opbeans-pipeline/). The reason for this appears to be that [the pipeline expects the existence of a Dockerfile](https://github.com/elastic/apm-pipeline-library/blob/master/.ci/dockerImagesOpbeans.groovy#L78-L87) and [none is found](https://github.com/elastic/observability-dev/issues/1511).

In looking at the way that the way that [the PHP Opbean is designed](https://github.com/elastic/opbeans-php/), it appears that there is a docker-compose file but no Dockerfile. In looking at [the docker-compose file](https://github.com/elastic/opbeans-php/blob/master/docker-compose.yml), it appears that the application is designed to run using [an off-the-shelf imagine which simply mounts the application](https://github.com/elastic/opbeans-php/blob/master/docker-compose.yml#L11-L21).

As such, we don't gain anything by pre-baking a Docker image because this has been done for us. Therefore, I believe we should remove this step which should resolve the failures in the pipeline.

## Related issues
Closes #ISSUE

### DIFF
--- a/.ci/dockerImagesOpbeans.groovy
+++ b/.ci/dockerImagesOpbeans.groovy
@@ -75,16 +75,6 @@ pipeline {
               push: true)
           }
         }
-        stage('Opbeans-php') {
-          agent { label 'docker' }
-          options { skipDefaultCheckout() }
-          steps {
-            buildDockerImage(repo: 'https://github.com/elastic/opbeans-php.git',
-              tag: "opbeans-php",
-              version: "${params.version}",
-              push: true)
-          }
-        }
         stage('Opbeans-python') {
           agent { label 'docker' }
           options { skipDefaultCheckout() }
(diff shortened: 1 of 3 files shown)


mdelapenya (approved):

cachedout:
Going to wait on @v1v to weigh in on this before merging.

kuisathaverat (approved):

v1v (approved):
That's a good point. The initial attempt was to use the same approach as other opbeans, I agree to revert to the previous configuration. 💯

PR #1105, opened 2021-05-10
Add resilience when the github API is down

v1v:
## What does this PR do?

A proposal to avoid any build failures when the GitHub API service is down, for such it adds a new argument called `ignoreGitHubFailures` (default true) that will avoid any build failures if the GitHub API call failed for whatever reason.

Downside:
- It might hide some other misconfiguration in the project though, but users might detect it in case the GitHub commit checks are not reported.

## Why is it important?

There were some GitHub API issues caused by the GitHub API service (see https://www.githubstatus.com/incidents/fm3pp6t2ytlh) and as a GitHub notify step in the CI pipelines failed

image

This is a proposal to avoid waste of time to debug failures which are caused by the third-party systems. 

## Questions

- What do you think?

### DIFF
--- a/vars/withGithubStatus.groovy
+++ b/vars/withGithubStatus.groovy
@@ -31,23 +31,37 @@ def call(Map args = [:], Closure body) {
   def description = args.get('description', context)
   def tab = args.get('tab', 'pipeline')
   def isBlueOcean = args.get('isBlueOcean', false)
+  def ignoreGitHubFailures = args.get('ignoreGitHubFailures', true)
 
   def redirect = detailsURL(tab: tab, isBlueOcean: isBlueOcean)
 
   try {
-    notify(context, "${description} ...", 'PENDING', redirect)
+    notifyMap(context: context, description: "${description} ...", status: 'PENDING', redirect: redirect, ignoreGitHubFailures: ignoreGitHubFailures)
     withAPM(){
       body()
     }
-    notify(context, "${description} passed", 'SUCCESS', redirect)
+    notifyMap(context: context, description: "${description} passed", status: 'SUCCESS', redirect: redirect, ignoreGitHubFailures: ignoreGitHubFailures)
   } catch (err) {
-    notify(context, "${description} failed", 'FAILURE', redirect)
+    notifyMap(context: context, description: "${description} failed", status: 'FAILURE', redirect: redirect, ignoreGitHubFailures: ignoreGitHubFailures)
     throw err
   }
 }
 
+// This is consumed by elastic/kibana.git@master:.ci/end2end.groovy
 def notify(String context, String description, String status, String redirect) {
+  notifyMap(context: context, description: description, status: status, redirect: redirect)
+}
+
+def notifyMap(Map args = [:]) {
   retryWithSleep(retries: 2, seconds: 5, backoff: true) {
-    githubNotify(context: "${context}", description: "${description}", status: "${status}", targetUrl: "${redirect}")
+    try {
+      githubNotify(context: "${args.context}", description: "${args.description}", status: "${args.status}", targetUrl: "${args.redirect}")
+    } catch (err) {
+      if (args.get('ignoreGitHubFailures', false)) {
+        log(level: 'WARN', text: "withGithubStatus: failed with error '${err.toString()}'. But 'ignoreGitHubFailures' has been enabled.")
+      } else {
+        throw err
+      }
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


kuisathaverat (approved):

cachedout:
Generally speaking I think things like this are good but I also think that we need to think about how we'll track the frequency of these events if we're ignoring them by default. I wonder if we could add a field to the results document that we send to the jenkins stats server which could contain these types of events. Then we could build some reporting based around that. WDYT?

(Shouldn't block this PR, necessarily, but it's just a thought.)

v1v:
> Generally speaking I think things like this are good but I also think that we need to think about how we'll track the frequency of these events if we're ignoring them by default. I wonder if we could add a field to the results document that we send to the jenkins stats server which could contain these types of events. Then we could build some reporting based around that. WDYT?

I like this idea, the only concern is the `notifyBuildStatus` and `githubNotifyStatus` steps stateless and not connected in any way. Therefore if we would like to gather those events we might need to think something else.

I initialised went to the distributed traces:

image

But even the build was marked with a failure I could not easily see what it could be the reason. I think, ideally, the distributed traces should provide those details.

Do you think if we move in that direction we could build some reporting with that?

vars/withGithubStatus.groovy
```diff
@@ -31,23 +31,41 @@ def call(Map args = [:], Closure body) {
...
+    notifyMap(context: context, description: "${description} failed", status: 'FAILURE', redirect: redirect, ignoreGitHubFailures: ignoreGitHubFailures)
     throw err
   }
 }
 
+// This is consumed by elastic/kibana.git@master:.ci/end2end.groovy
 def notify(String context, String description, String status, String redirect) {
-  retryWithSleep(retries: 2, seconds: 5, backoff: true) {
-    githubNotify(context: "${context}", description: "${description}", status: "${status}", targetUrl: "${redirect}")
+  notifyMap(context: context, description: description, status: status, redirect: redirect)
```
mdelapenya, line 52: Here the `ignoreGitHubFailures` is not passed so it's expected to pass the default value 'true', but in L64, the default value if not present is false. Is this on purpose?

v1v:
```suggestion
  // Fail in case the GitHub API calls fail for any reason
  notifyMap(context: context, description: description, status: status, redirect: redirect, ignoreGitHubFailures: false)
```

cachedout:
> I like this idea, the only concern is the notifyBuildStatus and githubNotifyStatus steps stateless and not connected in any way. Therefore if we would like to gather those events we might need to think something else.

Would something like a try/catch approach work where we can just catch the error, continue the job, but also pass the failure off to something that just creates an archive that has the failure?

Like I said previously, doesn't need to block this but I just think it's a pattern that could prove useful in the future if we could find a way to make it easy to implement.

mdelapenya (approved):

vars/withGithubStatus.groovy
```diff
@@ -31,23 +31,42 @@ def call(Map args = [:], Closure body) {
...
+  def i = 0
+  def numberOfRetries = 2
+  retryWithSleep(retries: numberOfRetries, seconds: 5, backoff: true) {
+    try {
+      i++
+      githubNotify(context: "${args.context}", description: "${args.description}", status: "${args.status}", targetUrl: "${args.redirect}")
+    } catch (err) {
+      // Let's retry at least as many times as possible otherwise, ignore the failure and avoid the notification in GitHub
+      if (args.get('ignoreGitHubFailures', false) && i == numberOfRetries) {
+        log(level: 'WARN', text: "withGithubStatus: failed with error '${err.toString()}'. But 'ignoreGitHubFailures' has been enabled.")
```
cachedout, line 66:
```suggestion
        log(level: 'WARN', text: "withGithubStatus: failed with error '${err.toString()}' but 'ignoreGitHubFailures' has been enabled.")
```

vars/withGithubStatus.txt
```diff
@@ -19,5 +19,6 @@ withGithubStatus(context: 'Release', tab: 'artifacts') {
 * description: Description of the GitHub status check. If unset then it will use the description.
 * tab: What kind of details links will be used. Enum type: tests, changes, artifacts, pipeline or an `<URL>`). Default pipeline.
 * isBlueOcean: Whether to use the BlueOcean URLs. Default `false`.
+* ignoreGitHubFailures: Whether to ignore when the github integration failed. Default `true`.
```
cachedout, line 22:
```suggestion
* ignoreGitHubFailures: Whether to ignore when the GitHub integration failed. Default `true`.
```

cachedout (approved):

v1v:
>Would something like a try/catch approach work where we can just catch the error, continue the job, but also pass the failure off to something that just creates an archive that has the failure?

I'll raise a GitHub issue about this discussion and then we can iterate on it

PR #1113, opened 2021-05-13
Remove windows versions used for the test-infra

v1v:
## What does this PR do?

`latest` images are the same as the other ones.

### DIFF
--- a/.ci/schedule-daily.groovy
+++ b/.ci/schedule-daily.groovy
@@ -39,6 +39,14 @@ pipeline {
   stages {
     stage('Run Tasks'){
       steps {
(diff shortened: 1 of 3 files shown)


kuisathaverat:
🤔 I use to check for tool changes in the daily builds if we change the frequency I could not. also if this pipeline blows up means everything is broken.

v1v:
> 🤔 I use to check for tool changes in the daily builds if we change the frequency I could not. also if this pipeline blows up means everything is broken.

No strong opinion, we can still run it manually if needed. So I'll keep the daily runs for now

cachedout (approved):

Issue #1121, opened 2021-05-21
Add GOARCH to withGoEnv step


mdelapenya:
@v1v do you think we can revert 0cd42e05badadaaa9e88c37ae4638033dcd3c8ff and apply that change to https://github.com/elastic/apm-pipeline-library/blob/6cfa9fac78ca050023ac0b5aa55de8d4090ff41c/vars/withGoEnvUnix.groovy#L48? Therefore, we will be setting the proper GOARCH value across any build.

_Originally posted by @mdelapenya in https://github.com/elastic/e2e-testing/pull/707#r589276064_

v1v:
I don't understand your question :/

What does the commit `0cd42e05badadaaa9e88c37ae4638033dcd3c8ff` refer to?

The `GOARCH` is configured in the contest of the installing the packages:

https://github.com/elastic/apm-pipeline-library/blob/b9cc99960eddbe64f85ea93715255b64247ea39a/vars/withGoEnvUnix.groovy#L67-L73

Do you think if moving that section to the L48 would be sufficient? I guess it should not cause any issues on the other hand.

v1v:
Aha, the comment in https://github.com/elastic/e2e-testing/pull/707/files#r589276064 :)

PR #1122, opened 2021-05-23
Set GOARCH env variable when using the withGoEnvUnix step

v1v:
## What does this PR do?

Set `GOARCH` on the fly based on the current Worker platform.

## Why is it important?

The consumers won't need to configure the GOARCH but simply use the step

## Related issues
Closes https://github.com/elastic/apm-pipeline-library/issues/1121

### DIFF
--- a/vars/withGoEnvUnix.groovy
+++ b/vars/withGoEnvUnix.groovy
@@ -41,11 +41,15 @@ def call(Map args = [:], Closure body) {
   def lastCoordinate = version[-2..-1]
   // gvm remove the last coordinate if it is 0
   def goDir = ".gvm/versions/go${lastCoordinate != ".0" ? version : version[0..-3]}.${os}.${arch}"
+  if (env.GOARCH?.trim() && env.GOARCH != arch) {
+    log(level: 'WARN', text: "withGoEnvUnix: GOARCH env variable matches '${env.GOARCH}' but it will be overridden to '${arch}'.")
+  }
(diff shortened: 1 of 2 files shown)


vars/withGoEnvUnix.groovy
```diff
@@ -41,11 +41,15 @@ def call(Map args = [:], Closure body) {
   def lastCoordinate = version[-2..-1]
   // gvm remove the last coordinate if it is 0
   def goDir = ".gvm/versions/go${lastCoordinate != ".0" ? version : version[0..-3]}.${os}.${arch}"
+  if (env.GOARCH?.trim() && env.GOARCH != arch) {
+    log(level: 'WARN', text: "withGoEnvUnix: GOARCH env variable matches '${env.GOARCH}' but it will be overridden to '${arch}'.")
```
mdelapenya, line 45: 👍 it matches supported values. Non-official list: https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63#a-list-of-valid-64-bit-goarch-values

mdelapenya (approved):

vars/withGoEnvUnix.groovy
v1v: Only to report when the GOARCH was configured in the consumer but it has got less precedence since it will be overridden anyway.

cachedout (approved):

PR #1127, opened 2021-05-24
chore: do not impact Elastic's download metrics

mdelapenya:
## What does this PR do?
It adds an URL param to Elastic's artifacts endpoint.

## Why is it important?
With this parameter, the download metrics won't be impacted by internal-automated hits.

### DIFF
--- a/resources/scripts/artifacts-api-latest-release-versions.sh
+++ b/resources/scripts/artifacts-api-latest-release-versions.sh
@@ -28,8 +28,8 @@
 set -eo pipefail
 
 URL="https://artifacts-api.elastic.co/v1"
-NO_KPI_URL_PARAMS="x-elastic-no-kpi=true"
+NO_KPI_URL_PARAM="x-elastic-no-kpi=true"
 
 OUTPUT=latest-release-versions.json
 
-curl -s "${URL}/versions?${NO_KPI_URL_PARAMS}" | jq -r '.versions[] | select(contains("SNAPSHOT")|not)' | jq -R . | jq -s '. | sort' | tee ${OUTPUT}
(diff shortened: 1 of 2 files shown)


resources/scripts/artifacts-api-latest-release-versions.sh
```diff
@@ -28,7 +28,8 @@
 set -eo pipefail
 
 URL="https://artifacts-api.elastic.co/v1"
+NO_KPI_URL_PARAMS="x-elastic-no-kpi=true"
 
 OUTPUT=latest-release-versions.json
 
-curl -s ${URL}/versions | jq -r '.versions[] | select(contains("SNAPSHOT")|not)' | jq -R . | jq -s '. | sort' | tee ${OUTPUT}
+curl -s "${URL}/versions?${NO_KPI_URL_PARAMS}" | jq -r '.versions[] | select(contains("SNAPSHOT")|not)' | jq -R . | jq -s '. | sort' | tee ${OUTPUT}
```
mdelapenya, line 35: I'm adding the parameter separator here instead of in the variable to make a clear separation on where the params do start

resources/scripts/artifacts-api-latest-release-versions.sh
```diff
@@ -28,7 +28,8 @@
 set -eo pipefail
 
 URL="https://artifacts-api.elastic.co/v1"
+NO_KPI_URL_PARAMS="x-elastic-no-kpi=true"
```
cachedout, line 31: Nit: this is a single parameter, so a better name might be `NO_KPI_URL_PARAM` (without the `S`). That will make it easier to read if we end up needing to add additional parameters.

cachedout (approved):

v1v (approved):

resources/scripts/artifacts-api-latest-release-versions.sh
mdelapenya: Yep, totally agree. Will change it ASAP

PR #1129, opened 2021-05-25
chore: add observer metadata for MB/FB

mdelapenya:
## What does this PR do?
It adds the observer metadata processor for metricbeat and filebeat

## Why is it important?
As explained in [docs](https://www.elastic.co/guide/en/beats/metricbeat/current/add-host-metadata.html), it recommends using `add_observer_metadata` if the beat is being used to monitor external systems. My question is: are Docker containers running in the host considered external systems?

### DIFF
--- a/resources/scripts/beats/filebeat.yml
+++ b/resources/scripts/beats/filebeat.yml
@@ -9,6 +9,7 @@ filebeat.autodiscover:
               - /var/lib/docker/containers/${data.docker.container.id}/*.log
 processors:
   - add_host_metadata: ~
+  - add_observer_metadata: ~
   - add_cloud_metadata: ~
   - add_docker_metadata: ~
   - add_kubernetes_metadata: ~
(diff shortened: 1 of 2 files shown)


v1v (approved):

v1v:
>  are Docker containers running in the host considered external systems

No idea :/

cachedout (approved):

mdelapenya:
> in this case `add_observer_metadata` and `add_host_metadata` because a Docker container has the Geolocation of his host, the same for k8s pods. So the option is not needed.

Yeah, I talked to @jsoriano yesterday afternoon, an he mentioned exactly that. The only difference would be duplicating fields under the `observer` key. My initial intention is to determine why there are no docs with `arm` architecture in our jenkins-stats cluster.

v1v:
As we are moving away form Jenkins specific implementations, I'll close this PR

Issue #1169, opened 2021-06-16
"googleStorageUploadExt" step should check if there are no matched files in the pattern


mdelapenya:
If that file does not exist, then `gsutil` fails

i.e.:

>[2021-06-16T14:56:28.509Z] + gsutil -m -q cp -a public-read build/system-tests-*.tar.gz gs://beats-ci-temp/Beats/beats/PR-26209-11
[2021-06-16T14:56:29.451Z] CommandException: No URLs matched: build/system-tests-*.tar.gz
[2021-06-16T14:56:31.359Z] CommandException: 1 file/object could not be transferred.
script returned exit code 1

v1v:
What do you want to this in this particular case? Patterns can be complex, so I'd rather prefer to keep the failure coming from `gsutil`

mdelapenya:
Yeah, I was thinking about that step at first instance too. I'd like to simply log that there are no matching files, and to skip the upload. This corner case has been found here: https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Fbeats/detail/PR-26209/11/pipeline/1245/, where the `args.upload` for the Beats seems to be true but there are not generated files to upload.

v1v:
I managed to find that we cannot use wildcards at the beginning of the pattern

PR #1193, opened 2021-07-04
Validate if pattern in googleStorageUploadExt returns any matches

v1v:
## What does this PR do?

Log if the pattern matches with any of the existing files

## Why is it important?

Ensure no failed stages in the pipeline but reporting with some log that it cannot happen.

## Related issues
Closes https://github.com/elastic/apm-pipeline-library/issues/1169

## Test

If no matching any pattern then it will notify with a warning message but won't fail.

image


v1v:
https://github.com/elastic/apm-pipeline-library/pull/1204 is the one that will validate if this PR is in a good shape

cachedout (approved):

kuisathaverat (approved):

PR #1197, opened 2021-07-06
Use a file to store the current and next release details

v1v:
## What does this PR do?

As long as `artifacts-api` does not support our needs, let's prepare prepare the system to rely on this repo to query what's the current release and the upcoming one for the 6.x and 7.x branches.

For such:
- a properties file with the current and next releases for the two minor active branches.
- a pipeline in charge to update the above file when needed. (to be defined the automation)
- updated the pipeline to read the above file instead using the `artifacts-api` entry point.
- `vars/bumpUtils.groovy` provides the public interface to read the above-mentioned properties file with some functions:
  - `getCurrentMinorReleaseFor7`
  - `getCurrentMinorReleaseFor6`
  - `getNextMinorReleaseFor7`
  - `getNextPatchReleaseFor7`

## Why is it important?

To ensure the contract with the consumers


.ci/updateStackReleaseVersion.groovy
```diff
@@ -0,0 +1,146 @@
...
+def createPullRequest(Map args = [:]) {
+  def message = createPRDescription(latestVersions)
+  prepareContext(repo: args.repo, branchName: args.branchName)
+  if (args?.stackVersions?.size() == 0) {
+    error('createPullRequest: stackVersions is empty. Review the artifacts-api for the branch ' + args.branchName)
+  }
+  writeFile file: 'resources/versions/releases.properties', text: """current_6=${releaseVersions.get('current.6')}
+current_7=${releaseVersions.get('current.7')}
+next_minor_7=${releaseVersions.get('next.minor.7')}
+next_patch_7=${releaseVersions.get('next.patch.7')}"""
```
kuisathaverat, line 84: to avoid having to change the configuration across reops every major, we can use something more neutral like:
* maintenance -> 6.x.x
* current -> 7.x.x
* current_next_minor -> 7.x.0
* current_next_patch -> 7.Y.x

v1v: I initially though to use something like that, but as long as the consumers use the same branch strategy (`6.<minor>`, `7.<minor>` or `7.x`) then the `current.7`/`current.6` could be easy to manage since the `<minor>` is explicitly in the branch name.

For instance:

- Given beats@7.x then I'd expect to use `next_minor_7` = 7.14.0

Therefore `next_minor_${branch.split('\.')[0]}` should be the way to know the value in the consumers

- Given beats@7.13 then I'd expect to use `next_patch_7` = 7.13.3

- Given beats@6.9 then I'd expect to use `current_6` = 6.8.15

I think the match will be easier when using the `<minor>` in the property name since there is no need to match other naming with the branch strategy.

What do you think?

PR #1217, opened 2021-07-22
feat: Otel pytest plugin

kuisathaverat:
## What does this PR do?

* Reimplementation of our pytest_apm plugin to work with OpenTelemetry

## Why is it important?

With this plugin we can report OpenTelemetry data related with the execution of tests implemented with pytest

## Test Locally

1. Run the OpenTelemetry Collector in debug mode
```shell
git clone git@github.com:kuisathaverat/apm-pipeline-library.git 
cd apm-pipeline-library/resources/scripts/pytest_otel
git checkout pytest_otel
make run-otel-collector
```
2. In another terminal in the same folder run the sample tests
```shell
make test
```
3. Check the OpenTelemetry Collector console output to see the transactions and span logs.
```shell
InstrumentationLibrary Test Suit 
Span #0
    Trace ID       : d2e52557b4865e2f3c19ec9e3f3c779b
    Parent ID      : 
    ID             : 200590773ff3bf30
    Name           : Running test_xfail
    Kind           : SPAN_KIND_INTERNAL
    Start time     : 2021-07-22 17:27:04.595932 +0000 UTC
    End time       : 2021-07-22 17:27:12.602354 +0000 UTC
    Status code    : STATUS_CODE_ERROR
    Status message : 
Attributes:
     -> test.name: STRING(test_xfail)
[...]

PR #1218, opened 2021-07-26
Provide OTEL context

v1v:
## What does this PR do?

Step to prepare the context in the production environment, for such it provides two masked env variables:
- `OTEL_EXPORTER_OTLP_ENDPOINT`
- `OTEL_TOKEN_ID`

Then the OTEL integration can use them safely.

## Why is it important?

https://github.com/elastic/apm-pipeline-library/pull/1217/ can benefit from and then we can start having one place where to look at.

## Important

I didn't add any further UTs that will require some mocking for:

- `Jenkins.get().getExtensionList`

## ITs

image


vars/isPluginInstalled.groovy
```diff
@@ -0,0 +1,26 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+  Given the pluginName it validates whether it's installed and available.
+
+*/
+def call(Map args = [:]) {
+  def pluginName = args.pluginName
+  def plugins = jenkins.model.Jenkins.instance.getPluginManager().getPlugins().collectEntries { [(it.getShortName()):it.getVersion()] }
```
cachedout, line 24: I have a question about this syntax. Is there a reason that `{ [(it.getShortName()):it.getVersion()] }` can't be `{ [it.getShortName():it.getVersion()] }` (Removing the first pair of parenthesis?)

cachedout (approved):

kuisathaverat (approved):

vars/isPluginInstalled.groovy
v1v: Syntax problems:

```
Script1.groovy: 1: illegal colon after argument expression;
   solution: a complex label expression before a colon must be parenthesized @ line 1, column 100.
   tEntries { [it.getShortName() :it.getVer
                                 ^
```

PR #1235, opened 2021-08-04
randomString without starting with -

v1v:
## What does this PR do?

Avoid a random string that starts with a `-`

## Why is it important?

Azure naming convention does not allow that

```
[2021-08-04T08:59:39.911Z] │ Error: "name" must begin with an alphanumeric character

```

### DIFF
--- a/vars/randomString.groovy
+++ b/vars/randomString.groovy
@@ -41,5 +41,5 @@ def call(Map args = [:]) {
   Alphanumeric and dash are allowed but not ending with dash
 */
 def manipulateString(String value, int size) {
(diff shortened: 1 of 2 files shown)


vars/randomString.groovy
```diff
@@ -41,5 +41,5 @@ def call(Map args = [:]) {
   Alphanumeric and dash are allowed but not ending with dash
```
mdelapenya, line 41:
```suggestion
  Alphanumeric and dash are allowed but not ending nor starting with dash
```

mdelapenya (approved):

PR #1246, opened 2021-08-23
Adjust weekly Beats headers to outsmart Google

ruflin:
The emails for the different branches are combined into a single email thread by Google which is confusing. Somehow Google assumes because there is only a slight change in
the prefix that it is the same mail thread (it is not). I'm trying to prevent this by adding the branch names also to the second part of the subject.


v1v (approved):
Thanks

cachedout (approved):

ruflin:
Is there a way to "trigger" an email run after this is merged to test it?

mdelapenya (approved):

PR #1248, opened 2021-08-24
findOldestSupportedVersion step

v1v:
## What does this PR do?

Step to find the oldest supported version of the stack given a version to compare with.

## Why is it important?

Simplify the pipelines

## Related issues

Caused by https://github.com/elastic/integrations/pull/1541

### DIFF
--- a/src/test/groovy/FindOldestStackVersionStepTests.groovy
+++ b/src/test/groovy/FindOldestStackVersionStepTests.groovy
@@ -1,0 +1,45 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import org.junit.Before
+import org.junit.Test
+import static org.junit.Assert.assertFalse
+import static org.junit.Assert.assertTrue
+
+class FindOldestStackVersionStepTests extends ApmBasePipelineTest {
+
+  @Override
+  @Before
+  void setUp() throws Exception {
+    super.setUp()
+    script = loadScript('vars/findOldestStackVersion.groovy')
+  }
+
+  @Test
+  void test_missing_argument() throws Exception {
+    testMissingArgument('versionCondition') {
+      script.call()
+    }
+  }
+
+  @Test
+  void test_no_match() throws Exception {
+    def ret = script.call(versionCondition: "^7.14.0")
+    printCallStack()
+  }
+
+}
--- a/src/test/groovy/FindOldestStackVersionStepTests.groovy
+++ b/src/test/groovy/FindOldestStackVersionStepTests.groovy
@@ -27,6 +27,27 @@ class FindOldestStackVersionStepTests extends ApmBasePipelineTest {
   void setUp() throws Exception {
     super.setUp()
     script = loadScript('vars/findOldestStackVersion.groovy')
+    helper.registerAllowedMethod('httpRequest', [Map.class], { f ->
+      return """{
+        "versions": [
+          "7.14.0-SNAPSHOT",
+          "7.14.0",
+          "7.14.1-SNAPSHOT",
+          "7.15.0-SNAPSHOT",
+          "7.15.0",
+          "8.0.0-SNAPSHOT"
+        ],
+        "aliases": [
+          "7.x-SNAPSHOT",
+          "7.14-SNAPSHOT",
+          "7.14",
+          "7.15-SNAPSHOT",
+          "7.15",
+          "8.0-SNAPSHOT",
+          "8.0"
+        ]
+      }"""
+    })
   }
 
   @Test
@@ -37,9 +58,23 @@ class FindOldestStackVersionStepTests extends ApmBasePipelineTest {
   }
 
   @Test
-  void test_no_match() throws Exception {
+  void test_match() throws Exception {
     def ret = script.call(versionCondition: "^7.14.0")
     printCallStack()
+    assertTrue(ret.equals('7.14.0'))
   }
 
+  @Test
+  void test_snapshot() throws Exception {
+    def ret = script.call(versionCondition: "^7.14.1")
+    printCallStack()
+    assertTrue(ret.equals('7.14.1-SNAPSHOT'))
+  }
+
+  @Test
+  void test_no_match() throws Exception {
+    def ret = script.call(versionCondition: "^7.13.0")
+    printCallStack()
+    assertTrue(ret.equals('7.13.0'))
+  }
 }
--- a/src/test/groovy/FindOldestSupportedVersionStepTests.groovy
+++ b/src/test/groovy/FindOldestSupportedVersionStepTests.groovy
@@ -20,13 +20,13 @@ import org.junit.Test
 import static org.junit.Assert.assertFalse
 import static org.junit.Assert.assertTrue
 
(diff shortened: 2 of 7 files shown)


vars/README.md
```diff
@@ -512,6 +512,20 @@ pipeline {
...
+## findOldestStackVersion
+Find the oldest stack version given the condition to compare with.
+
+If the version doesn't exist yet, it will try to use the closer snapshot, for example
+if 7.14.1 doesn't exist, it will try to use 7.14.1-SNAPSHOT or 7.x-SNAPSHOT,
+this will allow to develop integrations with unreleased features.
+
+
+```
+findOldestStackVersion(versionCondition: "^7.14.0")
```
cachedout, line 524: I'm finding the name of this step very hard to understand. What it seems like this is doing is trying to find the _newest_ image relative to the version that's passed in. However, `findOldestStackVersion` makes it seem like it's looking for the old _least new_ image. Am I not understanding the purpose of this step correctly?

src/test/groovy/FindOldestStackVersionStepTests.groovy
```diff
@@ -0,0 +1,80 @@
...
+    assertTrue(ret.equals('7.14.0'))
+  }
+
+  @Test
+  void test_snapshot() throws Exception {
+    def ret = script.call(versionCondition: "^7.14.1")
+    printCallStack()
+    assertTrue(ret.equals('7.14.1-SNAPSHOT'))
+  }
+
```
jsoriano, line 73: There is a corner case that happens during the FF of a release. There are artifacts for the version, the build candidates, but they haven't been released yet. In that case we want to use the snapshots, that are available. It'd be covered by a test case like this one:
```suggestion

  @Test
  void test_release_in_process() throws Exception {
    // There are already 7.15.0 artifacts, but this release hasn't happened yet, and there are only snapshots for the docker images.
    def ret = script.call(versionCondition: "^7.15.0")
    printCallStack()
    assertTrue(ret.equals('7.15.0-SNAPSHOT'))
  }
  
```

jsoriano (commented):
Nice, having testing coverage for this is great, thanks!

vars/README.md
jsoriano: The method is intended to find the oldest supported version for a given condition. Maybe `findOldestSupportedVersion` is better?

v1v: I agree, I'm not also onboarded with the naming, maybe `findCloserStackVersion` could be the correct one.

@jsoriano implemented a specific function in https://github.com/elastic/integrations/pull/1541/files#r694781463 , I C&C the content and added a few UTs. Let's wait for Jaime's feedback and see if it can be changed with something more specific.

cachedout: > Maybe findOldestSupportedVersion is better?

This makes a lot of sense. Big 👍 from me for this change.

src/test/groovy/FindOldestStackVersionStepTests.groovy
v1v: I think that's already done in https://github.com/elastic/apm-pipeline-library/pull/1248/files#diff-b9e4d7001a12a17bcee07878bce19ae78c6e8a0e520d53d59db0e141b87d73e1R68 with the `artifacts-api/.../versions` data structure:

```
{
        "versions": [
          "7.14.0-SNAPSHOT",
          "7.14.0",
          "7.14.1-SNAPSHOT",
          "7.15.0-SNAPSHOT",
          "7.15.0",
          "8.0.0-SNAPSHOT"
        ],
        "aliases": [
          "7.x-SNAPSHOT",
          "7.14-SNAPSHOT",
          "7.14",
          "7.15-SNAPSHOT",
          "7.15",
          "8.0-SNAPSHOT",
          "8.0"
        ]
      }
```

vars/findOldestStackVersion.groovy
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+Find the oldest stack version given the condition to compare with.
+
+def version = findOldestStackVersion(versionCondition: "^7.14.0")
+*/
+
+def call(Map args = [:]) {
+  def versionCondition = args.containsKey('versionCondition') ? args.versionCondition : error('findOldestStackVersion: versionCondition parameter is required')
+  def version = versionCondition.substring(1)
```
jsoriano, line 26: Current implementation only supports the `^` operator for versions. This is the only one used in integrations so far, maybe it should be mentioned in the docs here in case this function is used in other projects.

jsoriano: And maybe it would make sense to add something so `if versionCondition[0] != '^'`, it explicitly raises an exception or something saying that the condition is not supported?

src/test/groovy/FindOldestStackVersionStepTests.groovy
jsoriano: Notice that the cases are different:

* For `^7.14.0` there is a released version, but there are also snapshot artifacts, we want to test with the released version -> `7.14.0`.
* For `^7.14.1` we want to use `7.14.1-SNAPSHOT`, there are no `7.14.1` artifacts -> `7.14.1-SNAPSHOT`.
* For `^7.15.0` there is a released version and snapshot artifacts, but the version is not released yet and some artifacts as docker images are not available yet, we want to use the snapshots -> `7.15.0-SNAPSHOT`.

vars/findOldestStackVersion.groovy
v1v: See https://github.com/elastic/apm-pipeline-library/pull/1248/commits/001a757ad32ba33e035b80f812aeaad5574057fe

vars/README.md
v1v: See https://github.com/elastic/apm-pipeline-library/pull/1248/commits/7f20c354e38a7e72eb72e143b4dc9a5e2e93033e

src/test/groovy/FindOldestStackVersionStepTests.groovy
v1v: See https://github.com/elastic/apm-pipeline-library/pull/1248/commits/f06532f631b5b195d0fd1eb1c0c9ff509b15795d

Thanks Jaime!

mdelapenya (approved):

src/test/groovy/FindOldestSupportedVersionStepTests.groovy
```diff
@@ -0,0 +1,99 @@
...
+    assertTrue(assertMethodCallContainsPattern('error', 'versionCondition supports only'))
+  }
+
+  @Test
+  void test_release_in_process() throws Exception {
+    // There are already 7.15.0 artifacts, but this release hasn't happened yet, and there are only snapshots for the docker images.
+    def ret = script.call(versionCondition: "^7.15.0")
+    printCallStack()
+    assertTrue(ret.equals('7.15.0-SNAPSHOT'))
+  }
```
jsoriano, line 98: A couple of cases more that we could add:
```
  @Test
  void test_without_patch() throws Exception {
    def ret = script.call(versionCondition: "^7.14")
    printCallStack()
    assertTrue(ret.equals('7.14.0'))
  }

  @Test
  void test_next_minor() throws Exception {
    def ret = script.call(versionCondition: "^7.16.0")
    printCallStack()
    assertTrue(ret.equals('7.x-SNAPSHOT'))
  }
```

cachedout (approved):

src/test/groovy/FindOldestSupportedVersionStepTests.groovy
v1v:
```suggestion
  }
  
  @Test
  void test_without_patch() throws Exception {
    def ret = script.call(versionCondition: "^7.14")
    printCallStack()
    assertTrue(ret.equals('7.14.0'))
  }

  @Test
  void test_next_minor() throws Exception {
    def ret = script.call(versionCondition: "^7.16.0")
    printCallStack()
    assertTrue(ret.equals('7.x-SNAPSHOT'))
  }
```

v1v: I might need to change the logic in place to solve the issues with the UTs:

```
 groovy.lang.MissingPropertyException: No such property: availableVersion for class: findOldestSupportedVersion
```

Somehow the `for (availableVersion in availableVersions)` does not work correctly

v1v: > Somehow the for (availableVersion in availableVersions) does not work correctly

That was not the case but a wrong variable scope, so I just used the correct variable and things are back to normal :)

jsoriano (approved):
Thanks!

PR #1250, opened 2021-08-25
Add elastic specific env variables for withOtelEnv

v1v:
## What does this PR do?

Add elastic specific env variables then the consumers don't need to do anything but use the `withOtelEnv` instead of `withAPMEnv`.

## Why is it important?

Centralise traces in the same cluster.


mdelapenya (approved):
Awesome!

image

cachedout (approved):

PR #1255, opened 2021-09-03
Enable the opentelemetry

v1v:
## What does this PR do?

Enable the OpenTelemetry plugin that uses the ElasticStack we have in place.

https://github.com/elastic/apm-pipeline-library/pull/1051 was reverted since it created a local ElasticStack environment. So this approach uses our ES that runs in the Cloud.

## Why is it important?

To test this functionality locally and easily

## Related issues

This will help me to debug -> https://github.com/elastic/apm-agent-java/pull/2103 since somehow the `withOtelEnv` is not working as expected or something else is causing some issues.

## Caveats

You need to `make -C local clean` or destroy your local container.

Otherwise you might face issues with the existing opentelemetry version

## Test

image

### DIFF
--- a/local/configs/jenkins.yaml
+++ b/local/configs/jenkins.yaml
@@ -79,6 +79,11 @@ credentials:
               id: "apm-ci-gcs-plugin-file-credentials"
               scope: GLOBAL
               secretBytes: "${google_cloud_bucket_secret}" # secretBytes requires base64 encoded content
+          - string:
+              description: APM-Server Token
+              id: apm-server-token
+              scope: GLOBAL
+              secret: ${apmServerToken}
 
 unclassified:
   location:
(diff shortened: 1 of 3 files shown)


local/configs/jenkins.yaml
```diff
@@ -108,6 +113,22 @@ unclassified:
...
+      bearerTokenAuthentication:
+        tokenId: "apm-server-token"
+    endpoint: "${apmServerUrl}"
+    exportOtelConfigurationAsEnvironmentVariables: true
+    exporterIntervalMillis: 60000
+    exporterTimeoutMillis: 30000
+    ignoredSteps: "dir,echo,fileExists,isUnix,pwd,properties,readFile,readYaml,writeFile"
+    observabilityBackends:
+      - elastic:
+          kibanaBaseUrl: "${kibanaFriendlyUrl}"
```
v1v, line 127: I just added this entry in vault.

local/configs/plugins.txt
```diff
@@ -4,5 +4,6 @@ google-compute-engine
 job-dsl
 metrics
 monitoring
+opentelemetry
```
v1v, line 7: This is required to force the latest version of the plugin rather than using the existing one in the docker base image.

local/configs/jenkins.yaml
```diff
@@ -108,6 +113,22 @@ unclassified:
...
+          name: "Elastic Observability"
+    serviceName: "jenkins-local"
+    serviceNamespace: "jenkins-local"
```
v1v, line 130: Ensure the service is identified as a local environment

mdelapenya (approved):

kuisathaverat (approved):

PR #1260, opened 2021-09-06
Support traceparent with otel

v1v:
## What does this PR do?

Opentelemetry 0.19 already supports this feature, but let's support previous versions

## Why is it important?

Simplify the consumers of the withOtelEnv and other integrations to consume `TRACEPARENT` directly rather than generating it.

### DIFF
--- a/vars/withOtelEnv.groovy
+++ b/vars/withOtelEnv.groovy
@@ -47,6 +47,13 @@ def call(Map args = [:], Closure body) {
   withCredentials([string(credentialsId: credentialsId, variable: 'OTEL_TOKEN_ID')]) {
     def entrypoint = getEndpoint()
     def serviceName = getServiceName()
+
+    // Opentelemetry Jenkins plugin version 0.19 already provides the TRACEPARENT env
+    // variable, so let's support previous versions.
+    def otelEnvs = []
+    if (!env.TRACEPARENT) {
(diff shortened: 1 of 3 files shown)


vars/withOtelEnv.groovy
```diff
@@ -47,15 +45,24 @@ def call(Map args = [:], Closure body) {
   withCredentials([string(credentialsId: credentialsId, variable: 'OTEL_TOKEN_ID')]) {
     def entrypoint = getEndpoint()
     def serviceName = getServiceName()
+
+    // Opentelemetry Jenkins plugin version 0.19 already provides the TRACEPARENT env
+    // variable, so let's support previous versions.
+    def otelEnvs = []
+    if (!env.TRACEPARENT) {
```
kuisathaverat, line 52:
```suggestion
    if (!env.TRACEPARENT && env.TRACE_ID != null && env.SPAN_ID != null) {
```

kuisathaverat (approved):

vars/withOtelEnv.groovy
v1v: This won't be needed since https://github.com/elastic/apm-pipeline-library/pull/1260/files#diff-e2b84ca0a0839d4f9734edf9074e42dd41b772a19b07cdf4534bdf7ec36e8a12L32 verifies if the plugin is installed and therefore TRACE_ID and SPAN_ID should be there always.

PR #1261, opened 2021-09-06
Remove support for the OTEL service name env variable

v1v:
## What does this PR do?

Service Name should not be populated from the CI automatically

## Why is it important?

Service and version can clash:

image

image


mdelapenya (approved):

vars/withOtelEnv.groovy
```diff
@@ -55,9 +54,7 @@ def call(Map args = [:], Closure body) {
     withEnvMask(vars: [
       [var: 'ELASTIC_APM_SECRET_TOKEN', password: env.OTEL_TOKEN_ID],
       [var: 'ELASTIC_APM_SERVER_URL', password: entrypoint],
-      [var: 'ELASTIC_APM_SERVICE_NAME', password: serviceName],
```
mdelapenya, line 58: I guess that consumers should provide this env var, right?

v1v: That's right. IIRC, the e2e was one of the consumers, therefore it should work as usual

kuisathaverat (approved):

vars/withOtelEnv.groovy
v1v: For instance, https://github.com/elastic/e2e-testing/pull/1503/commits/0ecbe9c074969bbe4ca8debae5b9eb1dde5ec43a explicitly uses the ELASTIC_APM_SERVICE_NAME

PR #1264, opened 2021-09-07
Switch observability-dev checkouts from master branch by default to main branch

cachedout:
## What does this PR do?

Uses the main branch instead of the master branch when checking out `observability-dev`.

## Why is it important?

When using the `master` branch, we end up with an empty checkout. This causes failures such as https://apm-ci.elastic.co/job/apm-shared/job/bump-stack-version-pipeline/112/

### DIFF
--- a/.ci/bumpStackVersion.groovy
+++ b/.ci/bumpStackVersion.groovy
@@ -47,7 +47,7 @@ pipeline {
   stages {
     stage('Checkout') {
       steps {
-        git(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken', url: "https://github.com/${ORG_NAME}/${REPO}.git")
+        git(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken', url: "https://github.com/${ORG_NAME}/${REPO}.git", branch: "main")
       }
     }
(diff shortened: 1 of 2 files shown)


v1v (approved):

.ci/bumpStackVersion.groovy
```diff
@@ -47,7 +47,7 @@ pipeline {
   stages {
     stage('Checkout') {
       steps {
-        git(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken', url: "https://github.com/${ORG_NAME}/${REPO}.git")
+        git(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken', url: "https://github.com/${ORG_NAME}/${REPO}.git", branch: "main")
```
mdelapenya, line 50: Should this mean that any repo using this automation should have the `main` branch created?

cachedout: I don't think so. We're explicitly checking out `observability-dev` here: https://github.com/elastic/apm-pipeline-library/pull/1264/files#diff-eba3c46469d30c72b2ce74f5be13c5b7d85e69e0d8ef95dba6e5bf8e3f09a9e8R28

PR #1265, opened 2021-09-07
Beats webhook proof of concept

v1v:
## What does this PR do?

Use https://plugins.jenkins.io/generic-webhook-trigger/ to listen for events that will cause some actions

## Why is it important?

To iterate in the script in charge of the backport.

### DIFF
--- a/.ci/jobs/beats-webhook.yml
+++ b/.ci/jobs/beats-webhook.yml
@@ -12,7 +12,7 @@
         - git:
             url: git@github.com:elastic/apm-pipeline-library.git
             refspec: +refs/heads/*:refs/remotes/origin/* +refs/pull/*/head:refs/remotes/origin/pr/*
-            wipe-workspace: 'True'
+            wipe-workspace: true
             name: origin
             shallow-clone: true
             credentials-id: f6c7695a-671e-4f4f-a331-acdce44ff9ba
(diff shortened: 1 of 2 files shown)


kuisathaverat (approved):

.ci/jobs/beats-webhook.yml
```diff
@@ -0,0 +1,21 @@
...
+    concurrent: true
+    view: APM-CI
+    project-type: pipeline
+    pipeline-scm:
+      script-path: .ci/beats-webhook.groovy
+      scm:
+        - git:
+            url: git@github.com:elastic/apm-pipeline-library.git
+            refspec: +refs/heads/*:refs/remotes/origin/* +refs/pull/*/head:refs/remotes/origin/pr/*
+            wipe-workspace: 'True'
```
cachedout, line 15: Nit: It's a bit odd to see this value quoted. Is there a reason for that?

cachedout (approved):

.ci/jobs/beats-webhook.yml
v1v: Indeed, that's a C&P from https://github.com/elastic/apm-pipeline-library/blob/master/.ci/jobs/apm-schedule-daily.yml#L19

It should be safe to use a bool as far as I read in https://docs.openstack.org/infra/jenkins-job-builder/scm.html?highlight=wipe%20workspace#scm.git

Let me do it now

v1v:
```suggestion
            wipe-workspace: true
```

mdelapenya (approved):

PR #1276, opened 2021-09-15
fix: remove APM Server Docker images build

kuisathaverat:
## What does this PR do?

* remove APM Server Docker images build

## Why is it important?

It fixes the Docker images build https://apm-ci.elastic.co/blue/organizations/jenkins/apm-shared%2Fapm-docker-images-pipeline/detail/apm-docker-images-pipeline/1259/

### DIFF
--- a/.ci/buildDockerImages.groovy
+++ b/.ci/buildDockerImages.groovy
@@ -252,42 +252,6 @@ pipeline {
         )
       }
     }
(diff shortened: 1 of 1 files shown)


v1v (approved):
We don't use those images somewhere else?

kuisathaverat:
> We don't use those images somewhere else?

no, we removed the golang Dcoker image a few months ago.

PR #1277, opened 2021-09-15
Add nexus new implementation

v1v:
## What does this PR do?

Use  https://plugins.jenkins.io/apache-httpcomponents-client-4-api/ in `vars/nexusHelper.groovy`. Then we can deprecate `Nexus.groovy` that uses HttpURLConnection that produces some serialisation issues.

## Why is it important?

Fixes the serialisation issue with the `Nexus.groovy` implementation

```
an exception which occurred:
	in object org.jenkinsci.plugins.workflow.cps.CpsThreadGroup@562cc564
Caused: java.io.NotSerializableException: sun.net.www.protocol.https.HttpsURLConnectionImpl
```

## Related issues

Stage in the java agent is now manual -> https://github.com/elastic/apm-agent-java/pull/2127

### DIFF
--- a/vars/nexusHelper.groovy
+++ b/vars/nexusHelper.groovy
@@ -56,17 +56,18 @@ def getData(Map args = [:]) {
     // See https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/examples/org/apache/http/examples/client/ClientWithResponseHandler.java
     HttpGet httpget = new HttpGet(url)
     ResponseHandler<String> responseHandler = new ResponseHandler<String>() {
-        @Override
-        public String handleResponse(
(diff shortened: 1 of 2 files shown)


vars/nexusHelper.groovy
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.http.client.ClientProtocolException
+import org.apache.http.client.ResponseHandler
+
+import groovy.json.JsonSlurperClassic
+
+
+def createConnection(String host, String username, String password) {
+  CredentialsProvider credsProvider = new BasicCredentialsProvider()
+  credsProvider.setCredentials(
+          new AuthScope(host, 80),
```
SylvainJuge, line 38: Why not use `AuthScope.ANY_PORT` here ? Or parsing the port from the URL ?

PR #1278, opened 2021-09-15
GitHub comment bot commands

v1v:
## What does this PR do?

Add GH comment to re-trigger the build. 

## Why is it important?

Engage with the PR contributors to facilitate them how to interact with the CI

## Important

It only supports the pattern `^/test` for the time being, to avoid re-triggering builds over and over if a PR comment is added with the comment.

## UI

image

image

PR #1281, opened 2021-09-16
standardise the issueCommentTrigger with obltGitHubComments

v1v:
## What does this PR do?

Create a step with the default GitHub comments that are supported to trigger a build in the CI.

It adds `^` in the `Jenkins run the tests` comment and `^/test`.

## Why is it important?

To normalise the GitHub comments, then we can avoid hardcoded comments and have one place to rule out the same default behaviour in all the obs11 projects.

## Related issues

https://github.com/elastic/apm-pipeline-library/pull/1278

## Tests

These test sections reflects how this new step could be used in the pipelines were changed.

Comment https://github.com/elastic/apm-pipeline-library/pull/1281#issuecomment-921046689

Triggered

image

While https://github.com/elastic/apm-pipeline-library/pull/1281#issuecomment-921095119 didn't trigger any build, since it did not start with `Jenkins ....`


kuisathaverat:
Could we put that string in a step and use it in all the pipelines?


```
 triggers {
    issueCommentTrigger(commonFlags()+'|^/new-cmd$')
}
```

v1v:
This should not get triggered: Jenkins run the tests please

kuisathaverat (approved):

v1v:
@mdelapenya and @adam-stokes , I need your review here. is `obsGitHubComments` a good step name?

mdelapenya:
> @mdelapenya and @adam-stokes , I need your review here. is `obsGitHubComments` a good step name?

It's OK to me. I'd like to know if the `obs` prefix is going to be a standard for this kind of common steps. If that's not the case, let's look for a prefix that adds more context about the namespace, something like `oblt` or `o11y`...

mdelapenya (approved):

v1v:
> It's OK to me. I'd like to know if the obs prefix is going to be a standard for this kind of common steps. If that's not the case, let's look for a prefix that adds more context about the namespace, something like oblt or o11y...

@kuisathaverat @mdelapenya and @adam-stokes  can I ask you to vote for any of the below proposals within this comment?

😄 -> `obsGitHubComments`
👍 -> `obltGitHubComments`
❤️ -> `o11yGitHubComments`

adam-stokes:
+1 on the oblt prefix as that seems to be the standard naming scheme across projects

PR #1282, opened 2021-09-16
Support role and secret vault in dockerLogin

v1v:
## What does this PR do?

Add new parameters to the dockerLogin to configure the vault with different default values.

## Why is it important?

Allow to use different role/secret-id in Vault in a different CI

`dockerLogin` steps consumes [getVaultSecret](https://github.com/elastic/apm-pipeline-library/blob/master/vars/getVaultSecret.groovy) which already provides those parameters to be configured. Therefore, there is no need to set those parameters with any default values but pass through the `args` data structure from `dockerLogin` to `getVaultSecret`.

### DIFF
--- a/vars/dockerLogin.groovy
+++ b/vars/dockerLogin.groovy
@@ -25,8 +25,10 @@
 def call(Map args = [:]){
   def secret = args.containsKey('secret') ? args.secret : error("dockerLogin: No valid secret to looking for.")
   def registry = args.containsKey('registry') ? args.registry : "docker.io"
+  def role_id = args.containsKey('role_id') ? args.role_id : 'vault-role-id'
+  def secret_id = args.containsKey('secret_id') ? args.secret_id : 'vault-secret-id'
 
-  def jsonValue = getVaultSecret(secret: secret)
+  def jsonValue = getVaultSecret(secret: secret, role_id: role_id, secret_id: secret_id)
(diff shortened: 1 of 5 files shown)


kuisathaverat:
QQ the value of role/secret-id Is a credentials ID, isn't it?

v1v:
> QQ the value of role/secret-id Is a credentials ID, isn't it?

Exactly, see https://github.com/elastic/apm-pipeline-library/blob/fb193fcaa73991c1c3096bcd28cc484e0fe141da/vars/getVaultSecret.groovy#L66-L69

kuisathaverat (approved):

vars/dockerLogin.groovy
```diff
@@ -26,42 +26,36 @@ def call(Map args = [:]){
...
-            fi
-            if command -v dig 2>&1 > /dev/null; then
-              dig ${registry} 2>&1 > /dev/null 
-            fi
-            docker login -u "\${DOCKER_USER}" -p "\${DOCKER_PASSWORD}" "${registry}" 2>/dev/null
-            """)
-        } else {
-          bat(label: 'is registry service up?', script: """@ECHO OFF
-            nslookup ${registry} > NUL 2>&1
+  withEnvMask(vars: [
```
v1v, line 34: Use `withEnvMask`

PR #1285, opened 2021-09-20
[PoC] feat: add a CLI tool to generate the build data

mdelapenya:
## What does this PR do?
This is a PoC about simplifying the process to collect build information, and possibly sending it right after the build finishes. It mimics the behaviours described in the `generate-build-data.sh` shell script but using a programming language (Go), so that it's easier to test and follow the flow of the program.

In this PR, we are adding the `Jenkins Shipper` CLI tool as a Go command-line program, including:

- a `normaliser.go` file, including all functions that perform a sanitisation of the JSON data obtained from Jenkins. This file includes unit tests for all methods. The test methods will read from existing JSON test files under `src/test/resources`.
- a `json.go` file, with helper code to handle JSON objects: it's able to delete a key in a JSON object and to build a JSON object from an HTTP request.
- a `curl.go` file, to handle HTTP requests
- a `tools.go` file, to manage build tools dependencies, in particular `gotestsum` to run the tests with Make.
- a `main.go` file, including the main flow of the program. This main.go file will
	1. check for the existence of certain environment variables that are needed to proceed, in particular `BO_JOB_URL, BO_BUILD_URL, RESULT, DURATION, BUILD_URL`
	2. define a set of transformations and execute them one after the other. The transformations are:
		- prepareStepsInfo: will create the `steps-info.json` file
		- prepareTestErrors: will create the `tests-errors.json` file
		- preparePipelineLog: will create two files, `pipeline-log.txt` and `pipeline-log-summary.txt`
		- prepareBuildReport: will create the `build-report.json` file

Each transformation will perform an HTTP request, fetching from an URL, and converting the response to JSON (or an array of text lines in the case of the pipeline log transformation). The JSON object will be normalised using the right methods for each transformation.

## Why is it important?
1. Maintenance: current implementation is based in shell and `jq`. We are fetching from URLs converting the JSON/TXT responses into text files, using and reusing them across the script flow to build each report file. It's difficult to understand what file is used, as many times the target file is not an argument of the function, and internally used. 
2. Getting familiar with the code. Even though this PR does not get merged, I will be much more familiar with this shell script.

### Ubiquity
Shell is ubiquitous in the Linux Jenkins workers, therefore executing the script will always work on those machines. With this approach, we should provide 1) a Go runtime first, so that the Go program can be directly executed from the workspace; or 2) package the binary in a cross-platform manner so that each worker is able to receive the right binary and directly execute it.

### Performance
I've not measured the performance of running this Go code Vs running the shell script with the same amount of data. I can imagine shell being more performant in general, although the maintenance cost could help in mitigating it.

In that sense, and when discussing big number of elements, I can picture a situation where the number of tests is in the order of thousands (APM agents, in general). With the shell-script version, I've seen multiple calls to `jq`, constantly writing to disk. In the Go case, everything will happen in-memory. This is something to consider when accepting or not this PoC.

## How to test this?
There are a few JSON files under `src/test/resources` representing Jenkins responses. These files are used as inputs to be normalised by the unit tests of this project. You can run `make test` at the shipper directory to run the unit tests.

## Follow-ups
As suggested by @v1v, we could use a feature flag to cover the usage of this new model, using the shell script as fallback. I can imagine updating the `notifyBuildResult` step with something like:

```groovy
  if (env.GENERATE_BUILD_DATA_GO != '') {
    log(level: 'INFO', text: 'Using build data generation with Go')
    withGoEnv(){
      withEnv(["BO_JOB_URL=${restURLJob}, BO_BUILD_URL=${restURLBuild}, RESULT=${currentBuild.currentResult}, DURATION=${currentBuild.duration}"]){
        cmd(label: 'generate-build-data Go', returnStatus: true, script: 'go run main.go')
      }
    }
  } else {
    log(level: 'INFO', text: 'Using default build data generation with shell script')
    def scriptFile = 'generate-build-data.sh'
    def resourceContent = libraryResource("scripts/${scriptFile}")
    writeFile file: scriptFile, text: resourceContent
    sh(label: 'generate-build-data', returnStatus: true, script: """#!/bin/bash -x
      chmod 755 ${scriptFile}
      ./${scriptFile} ${restURLJob} ${restURLBuild} ${currentBuild.currentResult} ${currentBuild.duration}""")
  }
```

## Related issues
Closes #ISSUE


clis/jenkins-shipper/main.go
```diff
@@ -0,0 +1,509 @@
...
+			}
+
+			json = gabs.New()
+			json.Set(total, "total")
+			json.Set(passed, "passed")
+			json.Set(failed, "failed")
+			json.Set(skipped, "skipped")
+		}
+
+		return nil, err
```
v1v, line 379: IIUC, this manipulation not be returned here but later on?

```suggestion
```

mdelapenya: Mmm, you are right, but it's not a matter of removing the return there, but guarding it with the `else`

cachedout:
I think we should discuss a bit more about whether this is the right approach. 

I would like to understand in more depth why it's preferable to rebuild an already-functional script that's portable and working fine in production so far as I know. Could you please share some of the specific pain points of the existing script in more detail?

mdelapenya:
I think that, after our other initiatives to capture jUnit test-results, we can close this one. Thanks!

PR #1290, opened 2021-09-21
Add GitHub comment bot commands at the end

v1v:
## What does this PR do?

Add the GitHub commands list at the end of the comment

## Why is it important?

Usability

For instance, a build failure with some flakiness reporting might be too verbose and the command to re-trigger the build is not easyly accessible.

image

The proposal will do something like:

image

image

image


.ci/Jenkinsfile
```diff
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-@Library('apm@master') _
+@Library('apm@feature/add-github-comment-markdown') _
```
v1v, line 18:
```suggestion
@Library('apm@master') _
```

PR #1296, opened 2021-09-23
Remove test stats in the Flakyness github comment

v1v:
## What does this PR do?

Remove the test summary in the flaky comment.

## Why is it important?

It's already duplicated in the GitHub build comment which it runs always by default. So this could help to reduce the verbosity in the GitHub comments .

image


cachedout (approved):
Nice catch here.

PR #1298, opened 2021-09-23
Fix nodejs GitHub command

v1v:
## What does this PR do?

Fix GitHub command for the apm-agent-nodejs as stated in https://github.com/elastic/apm-agent-nodejs/pull/2345#pullrequestreview-762157365


trentm (approved):
Thanks.

PR #1300, opened 2021-09-27
fix: teach withGoEnv about Go executable package installation for Go 1.16 and up

### LINKED ISSUE #1299: withGoEnv is not aware of Go >=1.16 executable installation recipe changes
`withGoEnv` is used to install Go packages. The documentation does not make clear whether these are only executable packages or are any arbitrary package. Since Go 1.16, the approach used for installing executables has changed to use go install with a version suffix (see [here](https://go.dev/blog/go116-module-changes) under "Installing an executable at a specific version").

If `withGoEnv` is only used to install executables, then the fix is to conditionally use `go get` or `go install` based on the Go version. If it is used to install arbitrary packages, then the fix is more complicated, requiring the caller to indicate which type of package is being installed. From a brief survey of calls, happily it looks like only executables are being installed.

See https://github.com/elastic/stream/pull/27#issuecomment-927398831 for back ground.

efd6:
## What does this PR do?

This change fixes Go executable package installation for Go versions at and above Go version 1.16.

## Why is it important?

Currently, the installation method used by `withGoEnv` is `go get -u` which modifies go.mod and go.sum, causing subsequent tests to fail if building with a go1.16+ toolchain. The preferred method for installation of executable packages since Go 1.16 is `go install host.dom/path/to/pkg@version` with `@version` being `@latest` for the equivalent to the `-u` currently used with `go get` (see [here](https://go.dev/blog/go116-module-changes) under "Installing an executable at a specific version").

## Related issues
Fixes #1299

### DIFF
--- a/vars/isBeforeGo1_16.groovy
+++ b/vars/isBeforeGo1_16.groovy
@@ -1,0 +1,40 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+  If the given Golang version is pre 1.16.
+
+  whenTrue(isBeforeGo1_16(version: '1.17')) {
+    ...
+  }
+*/
+def call(Map args = [:]){
+  def version = args.containsKey('version') ? args.version : goDefaultVersion()
+  version = version.startsWith('go') ? version.minus('go') : version
+  def ref = "1.16"
+  def vVer = version.tokenize(".")
+  def vRef = ref.tokenize(".")
+  def n = Math.min(vVer.size(), vRef.size())
+  for (int i = 0; i < n; i++) {
+    def q = vVer[i].toInteger()
+    def r = vRef[i].toInteger()
+    if (q != r) {
+      return q < r
+    }
+  }
+  return vVer.size() < vRef.size()
+}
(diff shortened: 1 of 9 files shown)


v1v:
I'll made a couple of changes in this PR to reuse `isBeforeGo1_16`

vars/withGoEnvUnix.groovy
```diff
@@ -68,10 +68,13 @@ def installPackages(Map args = [:]) {
   // GOARCH is required to be able to install the given packages for the specific arch
   def arch = (env.GOARCH?.trim()) ?: goArch()
   log(level: 'DEBUG', text: "installPackages: GOARCH=${arch}")
+  def version = args.containsKey('version') ? args.version : goDefaultVersion()
+  def method = isBeforeGo1_16(version: version) ? "get -u" : "install"
+  def atVersion = isBeforeGo1_16(version: version) ? '' : "@latest"
```
v1v, line 73: Initially, there was no assigned value for a match, so I forced with ''

vars/withGoEnvWindows.groovy
```diff
@@ -46,6 +46,8 @@ def call(Map args = [:], Closure body) {
   def goDir = ".gvm\\versions\\go${lastCoordinate != ".0" ? version : version[0..-3]}.${os}.${goArch}"
   def goRoot = "${userProfile}\\${goDir}"
   def path = "${env.WORKSPACE}\\bin;${goRoot}\\bin;${chocoPath};C:\\tools\\mingw${mingwArch}\\bin;${env.PATH}"
+  def method = isBeforeGo1_16(version: version) ? "get -u" : "install"
+  def atVersion = isBeforeGo1_16(version: version) ? '' : "@latest"
```
v1v, line 50: Initially, there was no assigned value for a match, so I forced with ''

kuisathaverat (approved):

vars/isBeforeGo1_16.groovy
```diff
@@ -0,0 +1,40 @@
...
+/**
+  If the given Golang version is pre 1.16.
+
+  whenTrue(isBeforeGo1_16(version: '1.17')) {
+    ...
+  }
+*/
+def call(Map args = [:]){
+  def version = args.containsKey('version') ? args.version : goDefaultVersion()
+  version = version.startsWith('go') ? version.minus('go') : version
```
efd6, line 27: Note that I included this due to Go version string conventions. It may not be necessary here because all the handed in version strings lack the "go" prefix.

adam-stokes (approved):

mdelapenya (approved):

efd6:
I will leave this for you to merge. Thank you.

PR #1309, opened 2021-10-05
Retry side effect

cachedout:
## What does this PR do?

Allows the use of an optional closure to be run before each retry. This is useful for things like resetting the environment prior to a retry.

## Related issues
Refs https://github.com/elastic/apm-agent-nodejs/issues/2293

### DIFF
--- a/vars/retryWithSleep.groovy
+++ b/vars/retryWithSleep.groovy
@@ -32,6 +32,7 @@ def call(Map args = [:], Closure body) {
   def seconds = args.get('seconds', 10)
   def backoff = args.get('backoff', false)
   def sleepFirst = args.get('sleepFirst', false)
+  def sideEffect = args.get('sideEffect', { })
 
   def factor = 0
   retry(i) {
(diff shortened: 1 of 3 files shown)


vars/retryWithSleep.groovy
```diff
@@ -32,6 +32,7 @@ def call(Map args = [:], Closure body) {
   def seconds = args.get('seconds', 10)
   def backoff = args.get('backoff', false)
   def sleepFirst = args.get('sleepFirst', false)
+  def sideEffect = args.get('sideEffect', { })
```
mdelapenya, line 35: The word sideEffect usually has a negative connotation. What do you think about using `callback` instead?

mdelapenya: If that was your intention, I'm OK too :)

mdelapenya (approved):

vars/retryWithSleep.groovy
v1v: very tiny neat:

For those shared library steps that support closures, we use the signature `def call (Map args=[:], Closure body)` so maybe we could use the same class type for clarity?

```suggestion
  def sideEffect = args.get('sideEffect', Closure)
```

PR #1318, opened 2021-10-13
Use native support for concurrent builds kill

v1v:
**THIS SHOULD BE MERGED ONLY ONCE WE KNOW workflow-job-2.42 IS AVAILABLE**

## What does this PR do?

Enable the new option when possible and keep support for previous configuration

## Why is it important?

Use native support as implemented in https://github.com/jenkinsci/workflow-job-plugin/pull/200

## Related issues

Implementation https://github.com/jenkinsci/workflow-job-plugin/releases/tag/workflow-job-2.42

It's now available

image

And produced 

image


v1v:
It requires to search for the InterruptedAction:

image

To be able to use the same GitHub comment when builds are aborted for a new build

mdelapenya (approved):

cachedout (approved):

PR #1332, opened 2021-10-21
feat: add OS information for each build

mdelapenya:
## What does this PR do?
This PR adds a new build data file, `os-info.json`, containing the following information:

- OS name, obtained from Linux's `/etc/os-release`
- OS version, obtained from Linux's `/etc/os-release`
- Machine Architecture, obtained from Linux `uname -i`

The data will be added to a new JSON object, with `os` key, that will be appended to the already existing `build-report.json` file, so the new data will be as discoverable as the existing one.

## Why is it important?
This will provide information about where the build has been executed, important for test execution and OS/ARCH support. If/when stored in Elasticsearch, we could search for tests running on Linux > Ubuntu > x86_64.

## Other concerns
Because the `notifyBuildResult` step only works on Linux, these changes are only valid for Unix workers, not Windows.

## Related issues
Closes #ISSUE

### DIFF
--- a/resources/scripts/generate-build-data.sh
+++ b/resources/scripts/generate-build-data.sh
@@ -163,7 +163,6 @@ function fetchAndPrepareBuildReport() {
 function fetchAndPrepareOSReport() {
     file=$1
     key=$2
-    default=$3
 
     name=$(grep -E '^(NAME)=' /etc/os-release | awk -F\" '{print $(NF-1)}')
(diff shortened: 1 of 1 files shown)


mdelapenya:
@v1v mentions an important thing: this script is executed in the Jenkins master instance, so it's not valid for our purposes... 💩

mdelapenya:
Closing until we elaborate a better approach for multi-platform support. We can reopen if/when needed

PR #1346, opened 2021-11-02
Add test coverage for multi field arguments in gh

v1v:
## What does this PR do?

Add test coverage for the `gh` step to validate the same field can be set several times.

## Why is it important?

To illustrate this is possible

### DIFF
--- a/src/test/groovy/GhStepTests.groovy
+++ b/src/test/groovy/GhStepTests.groovy
@@ -182,4 +182,11 @@ class GhStepTests extends ApmBasePipelineTest {
     printCallStack()
     assertTrue(assertMethodCallContainsPattern('sh', """--search='"[automation] foo bar" in:title'"""))
   }
+
+  @Test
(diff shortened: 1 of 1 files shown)


nxei (approved):

nxei:
The test is name is not descriptive. Can you change it to something like this:

test_with_multiple_flag_entries

src/test/groovy/GhStepTests.groovy
```diff
@@ -182,4 +182,11 @@ class GhStepTests extends ApmBasePipelineTest {
     printCallStack()
     assertTrue(assertMethodCallContainsPattern('sh', """--search='"[automation] foo bar" in:title'"""))
   }
+
+  @Test
+  void test_with_same_field_twice() throws Exception {
```
v1v, line 187:
```suggestion
  void test_with_multiple_flag_entries() throws Exception {
```

PR #1353, opened 2021-11-03
Add gh version and forceInstallation paramters

v1v:
## What does this PR do?

Set the gh version to be installed and whether to force the installation regardless

## Why is it important?

GH output might differ between versions so let's help the consumers to deefine what version to be used.

### DIFF
--- a/vars/gh.groovy
+++ b/vars/gh.groovy
@@ -37,7 +37,7 @@ def call(Map args = [:]) {
   def command = args.containsKey('command') ? args.command : error('gh: command parameter is required.')
   def credentialsId = args.get('credentialsId', '2a9602aa-ab9f-4e52-baf3-b71ca88469c7')
   def flags = args.get('flags', [:])
-  def version = args.get('version', [:])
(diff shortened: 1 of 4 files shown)


vars/gh.groovy
```diff
@@ -87,8 +89,8 @@ def runCommand(command, flagsCommand) {
   return output
 }
 
-def downloadInstaller(where) {
-  def url = 'https://github.com/cli/cli/releases/download/v1.9.2/gh_1.9.2_linux_amd64.tar.gz'
+def downloadInstaller(where, version) {
+  def url = "https://github.com/cli/cli/releases/download/v${version}/gh_${version}_linux_amd64.tar.gz"
```
kuisathaverat, line 93: NIT: on macOS will fail to execute the binary.

kuisathaverat (approved):

vars/gh.groovy
v1v: I'll do a follow up

PR #1354, opened 2021-11-03
Pin the gh version for the listGithubReleases step

v1v:
## What does this PR do?

Pin the gh version for the listGithubReleases step

## Why is it important?

It parses the output, therefore let's ensure any format changes in the output don't affect the step


cachedout (approved):
💯 Nice one.

kuisathaverat (approved):

PR #1355, opened 2021-11-03
Support for macos, arm and windows in the gh step

v1v:
## What does this PR do?

Support for `Windows`, `Darwin` and `Arm`

## Why is it important?

`isUnix()` returns true for `Darwin` and `Arm`, therefore there was a chance that the consumers run the gh step in one of those OS/Arch, this should avoid that issue

Support for Windows in case it's needed...

### DIFF
--- a/src/test/groovy/GhStepTests.groovy
+++ b/src/test/groovy/GhStepTests.groovy
@@ -32,17 +32,22 @@ class GhStepTests extends ApmBasePipelineTest {
   }
 
   @Test
-  void test_windows() throws Exception {
-    testWindows() {
+  void test_without_args() throws Exception {
+    testMissingArgument('command') {
       script.call()
     }
   }
 
   @Test
-  void test_without_args() throws Exception {
-    testMissingArgument('command') {
-      script.call()
-    }
+  void test_windows() throws Exception {
+    helper.registerAllowedMethod('isUnix', [], { false })
+    script.call(command: 'issue list', flags: [ label: 'foo'])
+    printCallStack()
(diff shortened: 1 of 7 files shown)


vars/gh.groovy
```diff
@@ -85,12 +86,17 @@ def call(Map args = [:]) {
...
 def runCommand(command, flagsCommand) {
-  def output = sh(label: "gh ${command}", script: "gh ${command} ${flagsCommand}", returnStdout: true)
+  def output = cmd(label: "gh ${command}", script: "gh ${command} ${flagsCommand}", returnStdout: true)
   return output
 }
 
 def downloadInstaller(where, version) {
-  def url = "https://github.com/cli/cli/releases/download/v${version}/gh_${version}_linux_amd64.tar.gz"
+  def os = nodeOS()
+  if (os.equals('darwin')) {
```
v1v, line 95: Unfortunately `gh` uses `macOS` -> 

image

src/test/groovy/GhStepTests.groovy
```diff
@@ -87,7 +92,7 @@ class GhStepTests extends ApmBasePipelineTest {
     printCallStack()
     assertTrue(assertMethodCallContainsPattern('withEnv', 'PATH+GH'))
     assertTrue(assertMethodCallContainsPattern('sh', 'wget -q -O'))
-    assertTrue(assertMethodCallContainsPattern('sh', '1.9.2'))
+    assertTrue(assertMethodCallContainsPattern('sh', 'linux_amd64.tar.gz'))
```
v1v, line 95: avoid to validate the version but the default arch and os version

vars/gh.groovy
kuisathaverat: 🤦

kuisathaverat (approved):

PR #1359, opened 2021-11-04
Find oldest version with or conditions

jsoriano:
## What does this PR do?

* Add support for "or" conditions in `findOldestSupportedVersion`.
* Add support for ">=" operator.
* Tests use `assert` instead of `assertTrue` so they display the checked values in case of failure.

## Why is it important?

* We may need packages that support multiple majors, for that we need to support expressions such as `^7.16.0 || ^8.0.0`.

## Related issues

- Issue with these expressions found in https://github.com/elastic/integrations/pull/2122.

### DIFF
--- a/src/test/groovy/FindOldestSupportedVersionStepTests.groovy
+++ b/src/test/groovy/FindOldestSupportedVersionStepTests.groovy
@@ -61,28 +61,28 @@ class FindOldestSupportedVersionStepTests extends ApmBasePipelineTest {
   void test_match() throws Exception {
     def ret = script.call(versionCondition: "^7.14.0")
     printCallStack()
-    assertTrue(ret.equals('7.14.0'))
+    assert ret.equals('7.14.0')
   }
 
   @Test
   void test_match_ge() throws Exception {
     def ret = script.call(versionCondition: ">=7.14.0")
     printCallStack()
-    assertTrue(ret.equals('7.14.0'))
+    assert ret.equals('7.14.0')
   }
 
   @Test
   void test_snapshot() throws Exception {
     def ret = script.call(versionCondition: "^7.14.1")
     printCallStack()
-    assertTrue(ret.equals('7.14.1-SNAPSHOT'))
+    assert ret.equals('7.14.1-SNAPSHOT')
   }
 
   @Test
   void test_no_match() throws Exception {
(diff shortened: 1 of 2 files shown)


src/test/groovy/FindOldestSupportedVersionStepTests.groovy
```diff
@@ -61,21 +61,28 @@ class FindOldestSupportedVersionStepTests extends ApmBasePipelineTest {
...
+    printCallStack()
+    assert ret.equals('7.14.0')
   }
 
   @Test
   void test_snapshot() throws Exception {
     def ret = script.call(versionCondition: "^7.14.1")
     printCallStack()
-    assertTrue(ret.equals('7.14.1-SNAPSHOT'))
+    assert ret.equals('7.14.1-SNAPSHOT')
```
v1v, line 78: OOC, What's the reason for using `assert` rather than `assertTrue`?

v1v (approved):

mtojek (approved):

src/test/groovy/FindOldestSupportedVersionStepTests.groovy
jsoriano: `assert` shows the values implied in the assertion, while `assertTrue` shows only an empty error.

Error with `assertTrue` when the assertion fails:
```
[ERROR] FindOldestSupportedVersionStepTests.test_or_condition  Time elapsed: 1.943 s  <<< FAILURE!
java.lang.AssertionError
	at FindOldestSupportedVersionStepTests.test_or_condition(FindOldestSupportedVersionStepTests.groovy:118)
```
And when `assert` fails:
```
[ERROR] FindOldestSupportedVersionStepTests.test_or_condition_reversed  Time elapsed: 0.123 s  <<< FAILURE!
org.codehaus.groovy.runtime.powerassert.PowerAssertionError: 
assert ret.equals('7.14.1')
       |   |
       |   false
       7.14.0
	at FindOldestSupportedVersionStepTests.test_or_condition_reversed(FindOldestSupportedVersionStepTests.groovy:132)
```

v1v:
As soon as the CI passes then it will be merged automatically. We will proceed with a release afterwards

mdelapenya (approved):

PR #1360, opened 2021-11-04
[mergify][codeowners] Auto-assign and add reviewers with Codeowners

v1v:
## What does this PR do?

Use mergify to auto-assign the PRs to the PR creator and assign `Reviewers` with the Codeowners

## Why is it important?

Less manual actions

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,3 +9,6 @@
 
 # Beats
 /resources/approval-list/beats.yml @andresrc
+
+# Auto assign reviews
+* @elastic/observablt-robots @elastic/observablt-robots-contractors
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -9,3 +9,6 @@
 
 # Beats
 /resources/approval-list/beats.yml @andresrc
+
+# Auto assign reviews
+* @elastic/observablt-robots @elastic/observablt-robots-contractors
```
cachedout, line 14: Unrelated to this PR specifically, but FWIW I don't see why we should have two separate groups at all.

cachedout (approved):

.github/CODEOWNERS
v1v: Exactly :)

mdelapenya (approved):

PR #1362, opened 2021-11-04
Add sshexec helper

adam-stokes:
Signed-off-by: Adam Stokes <51892+adam-stokes@users.noreply.github.com>

## What does this PR do?

Adds ability invoke ssh and execute commands on remote systems

```
sshexec(connection: "user@1.2.3.4", sshPrivateKey: ${env.WORKSPACE}/ssh_id_rsa, cmd: """sudo bash /home/user/script.sh """)
```

## Why is it important?

Simplifies having to pass in enabling psuedo tty, ignore host key checking etc

## Related issues
Closes #ISSUE

### DIFF
--- a/vars/sshexec.groovy
+++ b/vars/sshexec.groovy
@@ -1,0 +1,29 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+ Wrapper to call ssh and run a command on remote host
+
+ sshexec(connection: "user@1.2.3.4", sshPrivateKey: ${env.WORKSPACE}/ssh_id_rsa, cmd: """sudo bash /home/user/script.sh """)
+*/
+def call(Map args = [:]) {
+  def connection = args.get('connection', false)
+  def sshPrivateKey = args.get('sshPrivateKey', "~/.ssh/id_rsa")
+  def cmd = args.get('cmd', false)
+
+  return sh("ssh -tt -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i ${sshPrivateKey} ${connection} -- '${cmd}'")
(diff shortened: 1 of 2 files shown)


vars/sshexec.groovy
```diff
@@ -0,0 +1,29 @@
...
+/**
+ Wrapper to call ssh and run a command on remote host
+
+ sshexec(connection: "user@1.2.3.4", sshPrivateKey: ${env.WORKSPACE}/ssh_id_rsa, cmd: """sudo bash /home/user/script.sh """)
+*/
+def call(Map args = [:]) {
+  def connection = args.get('connection', false)
+  def sshPrivateKey = args.get('sshPrivateKey', "~/.ssh/id_rsa")
+  def cmd = args.get('cmd', false)
+
```
v1v, line 27: Maybe a failure if using windows to avoid any misleading?

```
  if(!isUnix()){
    error('sshexec: windows is not supported yet.')
  }
```

vars/sshexec.txt
```diff
@@ -0,0 +1,7 @@
+Wrapper to run ssh commands on a remote system
+
+Example:
+
+```
+sshexec(connection: "user@1.2.3.4", sshPrivateKey: ${env.WORKSPACE}/ssh_id_rsa, cmd: """sudo bash /home/user/script.sh """)
+``
```
v1v, line 7: If the windows then add the section about tit's not supported:

```
_NOTE_: Windows is not supported yet.
```

vars/sshexec.groovy
```diff
@@ -0,0 +1,29 @@
...
+/**
+ Wrapper to call ssh and run a command on remote host
+
+ sshexec(connection: "user@1.2.3.4", sshPrivateKey: ${env.WORKSPACE}/ssh_id_rsa, cmd: """sudo bash /home/user/script.sh """)
+*/
+def call(Map args = [:]) {
+  def connection = args.get('connection', false)
+  def sshPrivateKey = args.get('sshPrivateKey', "~/.ssh/id_rsa")
+  def cmd = args.get('cmd', false)
```
v1v, line 26: Can you add some UTs? To validate the parameters are populated as expected?

v1v (approved):
NIT, UTs and Windows support, no a blocker

vars/sshexec.groovy
adam-stokes: Ah ok, wasn't sure the best practice to raise that exception, I'll update the code, thanks!

cachedout (commented):
This whole thing kind of makes me nervous, tbh. I am not sure it is the best idea to be opening remote shells from ephemeral workers (or from Jenkins controllers?) and I think we should talk more about the use case for this and see if we can't solve it another way.

vars/sshexec.groovy
kuisathaverat: if you change `sh` by `cmd` probably it works in Windows too. My only doubt is if `-o UserKnownHostsFile=/dev/null` is valid also in windows, I think so

v1v:
@adam-stokes , is this still relevant?

v1v:
I'll close this PR as stalled

PR #1368, opened 2021-11-08
feat: override junit step supporting calling a callback

mdelapenya:
## What does this PR do?
This PR adds a new step overriding the built-in `junit` step, so that it's possible we are able to add extra logic to be executed right after `junit`.

At this moment, there is no custom logic apart from the call to junit, but we could enrich this step to add pre-defined actions adding them right after `junit`. 

## Why is it important?
We'd like to start defining logic that must always be executed when handling junit test results, such as:

- parsing test results files and sending aggregations to a backend (Elasticseach)
- detecting the runtime environment where the tests are run (OS, arch) and send it to a backend (Elasticseach)
- ...

## Follow-ups
We already have a `junitAndStore` step, which could be refactored to reuse this new step.

## Other concerns
Addressed in comments: ~~Are you OK with adding the `callback` extension point, specially security-wise? Maybe it's not a good idea to let consumers to extend the `junit` step with their own functionality and simply keep our defaults. Thoughts?~~

### DIFF
--- a/vars/junit.groovy
+++ b/vars/junit.groovy
@@ -34,7 +34,7 @@ Wrap the junit built-in step to perform extra operations with the test reports.
     }
 */
 def call(Map args = [:]) {
-  def hasCallback = args.containsKey('callback') ? true : false
+  def hasCallback = args.containsKey('callback')
   def testResults = args.containsKey('testResults') ? args.testResults : error('junit: testResults parameter is required')
   def junitArgs = args.findAll { k,v -> !(k.equals('callback')) }
 
--- a/vars/junit.groovy
+++ b/vars/junit.groovy
@@ -25,7 +25,7 @@ Wrap the junit built-in step to perform extra operations with the test reports.
                 post {
                     always {
                         // JUnit on steroids
-                        junit(callback: closure(), ...)
+                        junit(...)
                     }
                 }
             }
@@ -34,17 +34,7 @@ Wrap the junit built-in step to perform extra operations with the test reports.
     }
 */
 def call(Map args = [:]) {
-  def hasCallback = args.containsKey('callback')
-  def junitArgs = args.findAll { k,v -> !(k.equals('callback')) }
+  junit(args)
 
-  junit(junitArgs)
-
-  if (!hasCallback) {
-      return
-  }
-
-  def callback = args.callback
-
-  log(level: 'DEBUG', text: "junit: Running callback operations after junit")
-  callback()
+  log(level: 'DEBUG', text: "junit: Running extra operations after junit")
 }
--- a/vars/junit.groovy
+++ b/vars/junit.groovy
@@ -35,7 +35,6 @@ Wrap the junit built-in step to perform extra operations with the test reports.
 */
 def call(Map args = [:]) {
   def hasCallback = args.containsKey('callback')
-  def testResults = args.containsKey('testResults') ? args.testResults : error('junit: testResults parameter is required')
   def junitArgs = args.findAll { k,v -> !(k.equals('callback')) }
 
   junit(junitArgs)
--- a/vars/junit.groovy
+++ b/vars/junit.groovy
@@ -1,0 +1,51 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+Wrap the junit built-in step to perform extra operations with the test reports.
+
+    pipeline {
+        ...
+        stages {
+            stage(...) {
+                post {
+                    always {
+                        // JUnit on steroids
+                        junit(callback: closure(), ...)
+                    }
+                }
+            }
+        }
+        ...
+    }
+*/
+def call(Map args = [:]) {
+  def hasCallback = args.containsKey('callback') ? true : false
+  def testResults = args.containsKey('testResults') ? args.testResults : error('junit: testResults parameter is required')
+  def junitArgs = args.findAll { k,v -> !(k.equals('callback')) }
+
+  junit(junitArgs)
+
+  if (!hasCallback) {
+      return
+  }
+
+  def callback = args.callback
+
+  log(level: 'DEBUG', text: "junit: Running callback operations after junit")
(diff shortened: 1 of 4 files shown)


vars/junit.groovy
```diff
@@ -0,0 +1,51 @@
...
+                    }
+                }
+            }
+        }
+        ...
+    }
+*/
+def call(Map args = [:]) {
+  def hasCallback = args.containsKey('callback') ? true : false
+  def testResults = args.containsKey('testResults') ? args.testResults : error('junit: testResults parameter is required')
```
v1v, line 38: I'd say to delegate this to the `junit` step.

```suggestion
```

vars/junit.groovy
```diff
@@ -0,0 +1,51 @@
...
+                        junit(callback: closure(), ...)
+                    }
+                }
+            }
+        }
+        ...
+    }
+*/
+def call(Map args = [:]) {
+  def hasCallback = args.containsKey('callback') ? true : false
```
v1v, line 37: does `containsKey` return a boolean?

```suggestion
  def hasCallback = args.containsKey('callback')
```

v1v (approved):
NIT comments

v1v:
> Are you OK with adding the callback extension point, specially security-wise? Maybe it's not a good idea to let consumers to extend the junit step with their own functionality and simply keep our defaults. Thoughts?


Consumers can run the snippet if they want in their pipelines so, I don't think it will be a problem

vars/junit.groovy
mdelapenya: I copy&paste from the other key, and then refactored for simplification, forgetting to update the ternary operator, thanks!

mdelapenya:
> Consumers can run the snippet if they want in their pipelines so, I don't think it will be a problem

I had in mind somebody trying to send a pipeline to any of our projects, sending malicious code in the callback, like trying to read a secret from jenkins, removing files, etc...

vars/junit.groovy
mdelapenya: I copy&paste from the other key, and then refactored for simplification, forgetting to update the ternary operator, thanks!

v1v:
> I had in mind somebody trying to send a pipeline to any of our projects, sending malicious code in the callback, like trying to read a secret from jenkins, removing files, etc...

`gitCheckout` step avoids running builds for non-elasticians. And that's the very first stage in all the current pipelines, do you think will be enough? Or do you want an extra validation?

vars/junit.groovy
mdelapenya: I copied this assertion from the `junitAndStore` https://github.com/elastic/apm-pipeline-library/blob/master/vars/junitAndStore.groovy#L43

mdelapenya:
> > I had in mind somebody trying to send a pipeline to any of our projects, sending malicious code in the callback, like trying to read a secret from jenkins, removing files, etc...
> 
> `gitCheckout` step avoids running builds for non-elasticians. And that's the very first stage in all the current pipelines, do you think will be enough? Or do you want an extra validation?

What if an attacker refactors the pipeline to use the git checkout, and then uses this callback?

vars/junit.groovy
v1v: `junitAndStore ` is a bit opinionated, since it uses `stash` internally. In this case, it's needed for the `stash`. 
While in here, It's not used outside of the `junit` scope

mdelapenya: Makes sense. Will remove this verification, as we can re-add it whenever we use it

mdelapenya:
> What if an attacker refactors the pipeline to use the basic git checkout, and then uses this callback?

Jenkinsfile changes from untrusted forks are ignored, so my fears are prevented

cachedout:
I don't think I quite understand the use case for this. 

If the closure must always be defined and passed in to the overriden junit step, why not just execute that function directly after the junit step in the pipeline itself? Could you explain a bit more about why this approach is advantageous?

mdelapenya:
> I don't think I quite understand the use case for this.
> 
> If the closure must always be defined and passed in to the overriden junit step, why not just execute that function directly after the junit step in the pipeline itself? Could you explain a bit more about why this approach is advantageous?

Sure, we discussed it already offline, but let me explain it here too:

@v1v:
>this junit is the step consumed by other steps? or how do you foresee it?
sorry, I meant, if you see the junit(callback: '' to be called by the consumers or by specific new steps?
I prefer the later, than the formal since it can encapsulate some logic that could be beneficial for others

@mdelapenya:
>well, I’d say both at the beginning, and once we have narrowed down the use cases, create the specific steps for them

@v1v:
>Maybe, junit override step can be the one driving those decisions internally, something like:
if (env.REPO_NAME.equals(‘beats’) {
   // gather junit details and transform them to ES data)

@mdelapenya:
I also prefer the later, having specific steps
I added the callback to support iterating in a pipeline: adding custom logic before converting it into a step
drawbacks? the explosion of possible steps could not be detected properly :shrug:

cachedout:
Thanks for the explanation.

However, I'm still not following the logic here. I see no difference between:

```
    pipeline {
        ...
        stages {
            stage(...) {
                post {
                    always {
                        // JUnit on steroids
                        junit(callback: the_closure(), ...)
                    }
                }
            }
        }
        ...
    }
```

and
```
    pipeline {
        ...
        stages {
            stage(...) {
                post {
                    always {
                        // JUnit on steroids
                        junit()
                        the_closure()
                    }
                }
            }
        }
        ...
    }
```

If, as @v1v mentions in the snippet above, there were some sort of decision logic happening inside the overriden junit step I would understand why this might be necessary but without that, I'm still not following.

Apologies for the back and forth but I just want to make sure I understand. :) Thanks!

kuisathaverat:
to add the callback to the junit steps you want to add, you'd have to edit those calls, so I do not see the point to overwrite the junit step. Also, as Mike said the callback is called after the execution of the junit so an additional step after, it'd make the same function. I can imagine a step that wraps junit and whatever you want to run, but IMHO the overwrite is not justified here, when you overwrite a step you have to have extra care of the flow control, due to your proposal will pass a callback function this makes really difficult to take care of all flow control problems and scale the right error. If the flow control is not accurate you will have weird errors in the pipeline when something fail, and they are really difficult to track to the overwritten step, because of that overwriting a step must be the last resort.

mdelapenya:
I may should opened this PR as draft :) My first intention was to commit a placeholder step that could be iterable from then on. That's why the callback support was added until the real implementations are found, those impl that should go as fixed blocks in this overwritten step.

mdelapenya:
>when you overwrite a step you have to have extra care of the flow control, due to your proposal will pass a callback function this makes really difficult to take care of all flow control problems and scale the right error. If the flow control is not accurate you will have weird errors in the pipeline when something fail, and they are really difficult to track to the overwritten step, because of that overwriting a step must be the last resort.

Mmm this is super interesting. I used `retryWithSleep`'s `sideEffect` attribute as an implementation example, so we may revisit this step after this. I understand that adding the callback will be "risky", so I'll remove it from this PR, leaving to a simple override, where we will be incrementally building new feature step-by-step (at least that's my intention)

kuisathaverat:
@mdelapenya Is it superseded by https://github.com/elastic/apm-pipeline-library/pull/1514?

mdelapenya:
Yes, let me close it and keep the work on #1514

PR #1371, opened 2021-11-11
[mergify] notify to the assignee if PR was not merged yet

v1v:
## What does this PR do?

Notify the backported PRs from mergify if they have not been merged.

## Why

To test https://github.com/elastic/e2e-testing/pull/1792 (this PR is a subset of the e2e-testing one)

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -17,6 +17,8 @@ pull_request_rules:
       - label=backport-to-7.x
     actions:
       backport:
+        assignees:
+          - "{{ author }}"
         branches:
           - 7.x
         title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):

.mergify.yml
```diff
@@ -48,3 +50,13 @@ pull_request_rules:
...
+  - name: notify the backport has not been merged yet
+    conditions:
+      - -merged
+      - -closed
+      - author=mergify[bot]
+      - check-success=CLA
+    actions:
+      comment:
+        message: |
+          This pull request has not been merged yet. Could you please review it @{{assignee}}? 🙏
```
cachedout, line 62: Nit: Should this says `review and merge it`?

v1v: Good catch, this was only for testing purposes :phrew! And the one that will benefit from uses it fine:
- https://github.com/elastic/e2e-testing/blob/aa56b1f99d6da71f454a16d0d8dd2ae606d43be9/.mergify.yml#L100-L111

I'll fix it anyway to avoid C&P issues

PR #1385, opened 2021-11-16
k8s: enable the pod definition

v1v:
## What does this PR do?

Use k8s agents as much as possible.
More parallel stages (it's now two times faster, from 15 mnts to 7 mints)

## Why is it important?

Validate if the `apm-ci` current k8s configuration works as expected

For such I took the snippet from https://github.com/jenkinsci/kubernetes-plugin/blob/852a03ab0215e584e8d607b4d1359f6324621cfe/src/main/resources/org/csanchez/jenkins/plugins/kubernetes/pipeline/samples/declarative.groovy#L3-L31

### Provisioning time is nearly 0

image

### Pipeline on a PR basis

image

### Pipeline for all the stages (but release)

image

### Before

image

## Proposal

Restructure the pipeline to support:
- k8s specific stages
- Linux (GCP) specific stages
- Windows (GCP) specific stages
- GitHub commands for certain stages and UI parameter to force the run of all the stages regardless.

## Follow up

- Validate the Release stage works in k8s 

## Further details

`defaultContainer 'shell'` in the top level agent definition for the declarative pipeline can cause some issues! See [this stackoverflow thread](https://stackoverflow.com/questions/60671321/when-i-use-kubernetes-as-my-agent-in-jenkins-it-will-not-allow-another-executor)

```
hudson.model.Hudson cannot be cast to org.csanchez.jenkins.plugins.kubernetes.KubernetesSlave
```

## Actions

- If we use k8s then we need to make some changes in the steps that run docker for instance. ✅ 
- Labels are a hard requirement -> https://github.com/elastic/apm-pipeline-library/issues/1443


cachedout:
As mentioned [here](https://github.com/elastic/observability-robots/issues/889#issuecomment-970573346) we may need to discuss holding off on merging this for a little while.

.ci/Jenkinsfile
```diff
@@ -18,7 +18,23 @@
...
+  containers:
+  - name: shell
+    image: ubuntu
+    command:
+    - sleep
+    args:
+    - infinity
+  '''
+      defaultContainer 'shell'
+    }
```
kuisathaverat, line 36: 🤔 I think we can replace this definition with a step like it is possible to make with whole pipelines. It is an idea to test in the k8s project. Something like this:

```
pipeline {
  agent {  ubuntuPod() }

...
```

ubuntuPod.groovy
```
call(){
    kubernetes {
      yaml '''
apiVersion: v1
kind: Pod
spec:
  containers:
  - name: shell
    image: ubuntu
    command:
    - sleep
    args:
    - infinity
  '''
      defaultContainer 'shell'
    }
```

kuisathaverat (approved):

v1v:
Interesting failure when using k8s regarding the location where the files are stored:

```
18:02:53  Also:   hudson.remoting.Channel$CallSiteStackTrace: Remote call to apm-shared-apm-pipeline-library-mbp-pr-1385-3-13md8-lnpgm-3x1x1
18:02:53  		at hudson.remoting.Channel.attachCallSiteStackTrace(Channel.java:1797)
18:02:53  		at hudson.remoting.UserRequest$ExceptionResponse.retrieve(UserRequest.java:356)
18:02:53  		at hudson.remoting.Channel.call(Channel.java:1001)
18:02:53  		at hudson.FilePath.act(FilePath.java:1166)
18:02:53  		at hudson.FilePath.act(FilePath.java:1155)
18:02:53  		at hudson.FilePath.write(FilePath.java:2346)
18:02:53  		at org.jenkinsci.plugins.workflow.steps.WriteFileStep$Execution.run(WriteFileStep.java:116)
18:02:53  		at org.jenkinsci.plugins.workflow.steps.WriteFileStep$Execution.run(WriteFileStep.java:99)
18:02:53  		at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:47)
18:02:53  		at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
18:02:53  		at java.util.concurrent.FutureTask.run(FutureTask.java:266)
18:02:53  		at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
18:02:53  		at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
18:02:53  java.nio.file.AccessDeniedException: /home/jenkins/agent/workspace/apm-pipeline-library-mbp_PR-1385/src/github.com/elastic/apm-pipeline-library/.git/fetch.log
18:02:53  	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:84)
```

git fetch is stored to avoid long console output since we rpelaced the default checkout with our `gitCheckout` and fetch is happening.

cachedout (approved):

.ci/Jenkinsfile
v1v: Using `yamlFile`

.ci/Jenkinsfile
```diff
@@ -18,19 +18,20 @@
 @Library('apm@master') _
 
 pipeline {
-  agent { label 'linux && immutable' }
+  agent { kubernetes { yamlFile 'resources/k8s/UbuntuPod.yml' } }
```
kuisathaverat, line 21:
```suggestion
  agent { kubernetes ubuntuPod() }
```

.ci/Jenkinsfile
```diff
@@ -258,3 +299,8 @@ pipeline {
     }
   }
 }
+
```
kuisathaverat, line 302:
```suggestion
def ubuntuPod(){
 return { yamlFile 'resources/k8s/UbuntuPod.yml' }
}
```

kuisathaverat: this implemented as step (e.g. ubuntuPod()) or substep(e.g. pods.ubuntu()) will make sure easy to control all the k8s agents

.ci/Jenkinsfile
kuisathaverat: see https://github.com/elastic/apm-pipeline-library/pull/1385/files#r768908766

.ci/Jenkinsfile
kuisathaverat: a Map can also make the thing (pods.get('ubuntu'))

pods.groovy
```
def get(name){
  return pods[name]
}
```

v1v: I rather prefer to use the standard approach to define the k8s agent, let me explain the reason:

The k8s definition will be stored in each github project, therefore the shared library step will be a quite opinionated approach, while I don't see the benefits of using it, but adding certain indirections to finally call a relative path to a pods definition stored in the project. 

```
agent { kubernetes { yamlFile 'resources/k8s/UbuntuPod.yml' } }
```

vs

```
agent { kubernetes ubuntuPod() }
```

even the first one is more verbose, it reflects what's the agent and where the definition is stored, while the later is somehow more descriptive but it hides where the pod definition is stored.

PR #1387, opened 2021-11-17
[mergify] verify the creation

v1v:
I don't know if the `updated` could be related to no just changes in the files but added/removed labels, github comments.

So let's also try with created

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -69,7 +69,7 @@ pull_request_rules:
       - -closed
       - author=mergify[bot]
       - check-success=CLA
-      - schedule=Mon-Fri 13:00-14:00[Europe/Paris]
+      - schedule=Mon-Fri 16:00-17:00[Europe/Paris]
       - created-at<1 days ago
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):

.mergify.yml
```diff
@@ -50,16 +50,29 @@ pull_request_rules:
...
         message: |
-          This pull request has not been merged yet. Could you please review and merge it @{{ assignee | join(', @') }}? 🙏
+          UPDATED This pull request has not been merged yet. Could you please review and merge it @{{ assignee | join(', @') }}? 🙏
+  - name: notify the backport has not been merged yet (created)
+    conditions:
+      - -merged
+      - -closed
+      - author=mergify[bot]
+      - check-success=CLA
+      - schedule=Mon-Fri 13:00-14:00[Europe/Paris]
```
v1v, line 72:
```suggestion
      - schedule=Mon-Fri 16:00-17:00[Europe/Paris]
```

PR #1397, opened 2021-11-22
Bump versions

v1v:
## What does this PR do?

Bump the versions in the pipeline in charge to update them


kuisathaverat (approved):

cachedout (approved):

v1v:
For the record, this file update is needed to allow the automation in place to do the right thing when a new release is cut.

There is still some work to be done in order to identify the current releases, since we have been told that the `artifacts-api` is not the right entrypoint but `GitHub releases`.

For the time being this is manual until we do the automation

PR #1403, opened 2021-11-22
Updated TestUtils.withEnvInterceptor to fix IndexOutOfBoundsException

nxei:
## What does this PR do?

It fixes 3 problems in TestUtils.withEnvInterceptor

1. IndexOutOfBoundsException when environment variable defined with empty value. Like this:
```
withEnv(["ORG_NAME=", "REPO_NAME="])
```
2. Incorrect value in variables after withEnv exited. Current implementation has different behavior in comparison with how withEnv works in Jenkins. When withEnv  finished variables which was defined in argument string are set to null while in Jenkins they restored to values they had before withEnv  called.

3. If value for specified variable has = character it will have only value before this = character. This will set VAR to SOME instead of SOME=TEXT.
```
withEnv(["VAR=SOME=TEXT"]) 
```
## Why is it important?

It will allow use withEnv with empty variables value in unit tests. And withEnv in tests will work close to how it works in Jenkins.

### DIFF
--- a/src/test/groovy/co/elastic/TestUtils.groovy
+++ b/src/test/groovy/co/elastic/TestUtils.groovy
@@ -20,14 +20,18 @@ package co.elastic
 public class TestUtils {
 
   public static withEnvInterceptor = { list, closure ->
-    list.forEach {
-      def fields = it.split("=")
-      binding.setVariable(fields[0], fields[1])
+    def savedVars = []
+    // filter invlaid entries - with name empty or containing only spaces
+    list.findAll{ it&&it=~/^\s*[^=\s]+\s*=?/ }.forEach{
+      List fields = it.split("=");
+      def name = fields.remove(0);
+      def value = fields.join("="); // add back = chars deleted by split
+      savedVars.add([name, binding.hasVariable(name) ? binding.getVariable(name) : null])
+      binding.setVariable(name, value.size()> 0 ? value : null)
     }
     def res = closure.call()
-    list.forEach {
-      def fields = it.split("=")
-      binding.setVariable(fields[0], null)
+    savedVars.reverse().forEach {
+      binding.setVariable(it[0], it[1])
     }
     return res
   }
--- a/src/test/groovy/co/elastic/TestUtils.groovy
+++ b/src/test/groovy/co/elastic/TestUtils.groovy
@@ -21,7 +21,7 @@ public class TestUtils {
 
   public static withEnvInterceptor = { list, closure ->
     def savedVars = []
-    // filter invlaid entries - with name empty or containing only spaces
+    // filter invalid entries - with name empty or containing only spaces
     list.findAll{ it&&it=~/^\s*[^=\s]+\s*=?/ }.forEach{
       List fields = it.split("=");
       def name = fields.remove(0);


src/test/groovy/co/elastic/TestUtils.groovy
```diff
@@ -20,14 +20,18 @@ package co.elastic
...
+      def value = fields.join("="); // add back = chars deleted by split
+      savedVars.add([name, binding.hasVariable(name) ? binding.getVariable(name) : null])
+      binding.setVariable(name, value.size()> 0 ? value : null)
     }
     def res = closure.call()
-    list.forEach {
-      def fields = it.split("=")
-      binding.setVariable(fields[0], null)
+    savedVars.reverse().forEach {
+      binding.setVariable(it[0], it[1])
```
v1v, line 34: What's the reason for this particular change?

v1v:
> IndexOutOfBoundsException when environment variable defined with empty value. 
withEnv(["ORG_NAME=", "REPO_NAME="])

What's the reason to unset those variables? Where does it happen?

> Current implementation has different behavior in comparison with how withEnv works in Jenkins

Exactly, the existing implementation is not replacing Jenkins but helping with the UTs. So If needed it can be added to help w

> If value for specified variable has = character it will have only value before this = character. This will set VAR to SOME instead of SOME=TEXT.

Where do we use this particular `var=foo=value` format? If it's not used then I'd say we don't need to support it for the shake of simplicity.

src/test/groovy/co/elastic/TestUtils.groovy
nxei: The reason for this is to correctly restore variable value if this variable specified few times in arguments to withEnv.
LIke this:
```
withEnv(["VAR1=AAA","VAR2=BBB","VAR1=CCC"]) 
```
May be not a big improvement bit it was easy to do and Jenkins also works this way.

src/test/groovy/co/elastic/TestUtils.groovy
```diff
@@ -20,14 +20,18 @@ package co.elastic
 public class TestUtils {
 
   public static withEnvInterceptor = { list, closure ->
-    list.forEach {
-      def fields = it.split("=")
-      binding.setVariable(fields[0], fields[1])
+    def savedVars = []
+    // filter invlaid entries - with name empty or containing only spaces
```
v1v, line 24:
```suggestion
    // filter invalid entries - with name empty or containing only spaces
```

src/test/groovy/co/elastic/TestUtils.groovy
v1v: Gotcha! Thanks for the clarification

v1v (changes requested):
Please see my previous comment with some questions -> https://github.com/elastic/apm-pipeline-library/pull/1403#issuecomment-976308365

src/test/groovy/co/elastic/TestUtils.groovy
nxei: fixed

nxei:
Reply to [comment](https://github.com/elastic/apm-pipeline-library/pull/1403#issuecomment-976308365)

>> IndexOutOfBoundsException when environment variable defined with empty value. 
>withEnv(["ORG_NAME=", "REPO_NAME="])

>What's the reason to unset those variables? Where does it happen?

I tried to use it to run _gh_ step for _gh api_ call. Gh api dosen't accept  _--repo_ flag, So I tried to use it to get gh step running for _gh api_  and found that unit tests aborted with exception in this case.

>> Current implementation has different behavior in comparison with how withEnv works in Jenkins

>Exactly, the existing implementation is not replacing Jenkins but helping with the UTs. So If needed it can be added to help w

Agree with it, but if withEnv will work close to Jenkins it will be better because in future someone  can stumble on it like me. And also it cost nothing because it achieved while fixed issue with exceptions when set empty variables.

>> If value for specified variable has = character it will have only value before this = character. This will set VAR to SOME instead of SOME=TEXT.

>Where do we use this particular `var=foo=value` format? If it's not used then I'd say we don't need to support it for the shake of simplicity.

I think format  `var=foo=value` not in use anywhere, but mine idea here is as in the answer to the previous question - it can be used in Jenkins so why not support it?

v1v:
Thanks for the clarification. If If understand correctly this is not anymore a blocker but a nice to have feature, right?

I'm not fully onboard with this particular change since there are no UTs, and adding them would be a great contribution to this PR, my main concern is the `regex` pattern. I don't know what others might think about this.

nxei:
> Thanks for the clarification. If If understand correctly this is not anymore a blocker but a nice to have feature, right?

> I'm not fully onboard with this particular change since there are no UTs, and adding them would be a great contribution to this PR, my main concern is the regex pattern. I don't know what others might think about this.

I agree it's not blocker so should be changed to enhancement. Also about regex - it should be removed, because silently skip some vars is not good idea. I'll  fix it and also create unit tests.

v1v:
This PR is stalled, so I'll close it for now

PR #1406, opened 2021-11-22
Removed check for repo argument when specified forceRepo: true.

nxei:
## What does this PR do?

Removed check for repo argument when specified forceRepo: true.

## Why is it important?

This will allow to run gh command without --repo flag.
This is important because 'gh api' doesn't accept --repo flag

## Related issues
Related to: Allow repo customisation for the GH #1394

### DIFF
--- a/vars/gh.groovy
+++ b/vars/gh.groovy
@@ -45,7 +45,7 @@ def call(Map args = [:]) {
     log(level: 'DEBUG', text: 'gh: running within a git workspace.')
   } else {
     if (forceRepo) {
-      if (env.REPO_NAME?.trim() && env.ORG_NAME?.trim() && flags.containsKey('repo')) {
+      if (env.REPO_NAME?.trim() && env.ORG_NAME?.trim()) {
         log(level: 'WARN', text: 'gh: repo flag has higher precedence over REPO_NAME and ORG_NAME')
       }
     } else {
--- a/src/test/groovy/GhStepTests.groovy
+++ b/src/test/groovy/GhStepTests.groovy
@@ -247,4 +247,19 @@ class GhStepTests extends ApmBasePipelineTest {
     assertTrue(assertMethodCallContainsPattern('sh', "--repo='acme/bar'"))
     assertJobStatusSuccess()
   }
+
+  @Test
+  void test_outside_of_a_repo_with_variables_and_force_repo_wothout_repo_flag() throws Exception {
+    env.REPO_NAME = 'foo'
+    env.ORG_NAME = 'org'
+    helper.registerAllowedMethod('sh', [Map.class], { m ->
+      if (m?.returnStatus) { return 1 }})
+    try {
+      script.call(command: 'issue list', forceRepo: true)
+    } catch(err) { println err}
+    printCallStack()
+    assertFalse(assertMethodCallContainsPattern('sh', "--repo"))
+    assertJobStatusSuccess()
+  }
+
 }


vars/gh.groovy
```diff
@@ -45,7 +45,7 @@ def call(Map args = [:]) {
     log(level: 'DEBUG', text: 'gh: running within a git workspace.')
   } else {
     if (forceRepo) {
-      if (env.REPO_NAME?.trim() && env.ORG_NAME?.trim() && flags.containsKey('repo')) {
+      if (env.REPO_NAME?.trim() && env.ORG_NAME?.trim()) {
```
v1v, line 48: This condition was only enabled to report that `repo` flag in conjunction with `REPO_NAME` and `ORG_NAME` will always have higher precendence.

In other words:

```
withEnv(["REPO_NAME=foo", "ORG_NAME=acme"]) {
  gh(command: 'issue list', flags [ repo: 'elastic/my-repo' ])
}
```

Will use `--repo elastic/my-repo` rather than `--repo acme/foo`

Can you please elaborate what's the issue?

v1v: I guess in your case:

```
withEnv(["REPO_NAME=foo", "ORG_NAME=acme"]) {
  gh(command: 'api ....', forceRepo: true)
}
```

should work and `--repo acme/foo` which it's caused by the two environment variables should not be shown since `forceRepo` is set to true.

nxei: The reason for it is that I need run *gh api* command which fails if --repo argument is specified so I need to run it strictly without --repo flag.

```
user@lws7:~/tmp/elastic/gotest$ gh api --repo=aaa/bbb
unknown flag: --repo

Usage:  gh api <endpoint> [flags]

Flags:
      --cache duration        Cache the response, e.g. "3600s", "60m", "1h"
  -F, --field key=value       Add a typed parameter in key=value format
  -H, --header key:value      Add a HTTP request header in key:value format
      --hostname string       The GitHub hostname for the request (default "github.com")
  -i, --include               Include HTTP response headers in the output
      --input file            The file to use as body for the HTTP request (use "-" to read from standard input)
  -q, --jq string             Query to select values from the response using jq syntax
  -X, --method string         The HTTP method for the request (default "GET")
      --paginate              Make additional HTTP requests to fetch all pages of results
  -p, --preview strings       Opt into GitHub API previews
  -f, --raw-field key=value   Add a string parameter in key=value format
      --silent                Do not print the response body
  -t, --template string       Format the response using a Go template
```  

I could run it this way:
```
    withEnv(["ORG_NAME=", "REPO_NAME="]) {
        result = toJSON(gh(ghDefaultArgs(args) + [forceInstallation: true,
            command: "api repos/${repo}/actions/runs/${args.runId}"]))
    }
```
But for now *withEnv(["ORG_NAME=", "REPO_NAME="])*  causes exception in unit tests. I created another PR to fix it.

v1v: As I said in my previous comment, if you use `forceRepo: true` then the environment variables are ignored, to illustrate this please see the below UTs:

```
  @Test
  void test_gh_api_with_forceRepo_then_no_repo() throws Exception {
    env.REPO_NAME = 'foo'
    env.ORG_NAME = 'org'
    script.call(command: 'api repos/my-repo/actions/runs/my-run-id', forceRepo: true)
    printCallStack()
    assertFalse(assertMethodCallContainsPattern('sh', "--repo='org/foo'"))
    assertJobStatusSuccess()
  }

  @Test
  void test_gh_api_without_forceRepo_then_repo() throws Exception {
    env.REPO_NAME = 'foo'
    env.ORG_NAME = 'org'
    script.call(command: 'api repos/my-repo/actions/runs/my-run-id', forceRepo: false)
    printCallStack()
    assertTrue(assertMethodCallContainsPattern('sh', "--repo='org/foo'"))
    assertJobStatusSuccess()
  }
```

nxei: agree

PR #1409, opened 2021-11-23
Gha build get workflow run forced repo

nxei:
## What does this PR do?

It fixes call githubWorkflowRun.getWorkflowRun when set environment variables env.ORG_NAME and env.REPO_NAME.


v1v (changes requested):
1) Can you please fix the commit linting defect? Those defects can be detected locally with the `pre-commit`, the installation details can be found in https://github.com/elastic/apm-pipeline-library/blob/master/docs/PRECOMMIT.md#installation


Then every time there is a git commit the pre-commit will evaluate the existing rules and will fail if a defect is detected.

The good thing is some of the linting rules do fix those defects so, you only need to git add and git commit again.

2) Do you think a new Unit test case could help to document this requirement? If so, please add one.

src/test/groovy/GithubWorkflowRunTests.groovy
```diff
@@ -259,4 +259,36 @@ View this run on GitHub: https://github.com/owner/repo/actions/runs/1441316856
...
+
+  @Test
+  void test_getWorkflowRun_works_properly_with_ORG_NAME_and_REPO_NAME_set() throws Exception {
+    env.REPO_NAME = 'foo'
+    env.ORG_NAME = 'org'
+    def result = script.getWorkflowRun(runId: 1441316856, repo: "owner/repo")
+    def expected = [id: 1441316856, status: "completed"]
+    assertTrue(expected.id == result.id && expected.status == result.status)
+    env.REPO_NAME = 'repo'
+    env.ORG_NAME = 'owner'
```
v1v, line 271: Is this needed? IIRC, if `env.NAME` is defined is already set, so  L265 and L266 already set the values.

nxei: yes, it's important because it checks different case - when repo argument is not specified and  repo info is taken from these environment variables

v1v: I don't see any assertion for those particular environment variables, therefore I don't see why we need them twice.
If they are needed, can you add the required assertion?

v1v (changes requested):
Please read the previous comment regarding the `pre-commit`, I have not seen any new changes to fix the linted error.

If you run the below command locally:

```
pre-commit run -v --files vars/githubWorkflowRun.groovy
```

You can see

```
Trim Trailing Whitespace.................................................Failed
- hook id: trailing-whitespace
- duration: 0.07s
- exit code: 1
- files were modified by this hook

Fixing vars/githubWorkflowRun.groovy
```

So, you can then git add and git commit the changes

Thanks

src/test/groovy/GithubWorkflowRunTests.groovy
nxei: Values from these variables used to create gh api command line. Then this command like checked in  *ghClosure* closure where encoded only 3 cases for testing *gh api* calls. All these cases are for repo value *owner/repo*. So to use one of  them for checking this case I need to set these variables to 
```
    env.REPO_NAME = 'repo'
    env.ORG_NAME = 'owner'
```
This is all cases which currently are in ghClosure:
```
      if (args.command == "api repos/owner/repo/actions/runs/1441316856") {
        return '{"id": 1441316856, "status": "completed"}'
      }
      if (args.command == "api repos/owner/repo/actions/runs/1441334029") {
        return '{"message": "Not Found"}'
      }
      if (args.command == "api repos/owner/repo/actions/runs/1441316123") {
        return '{"id": 1441316123, "status": "queued"}'
      }
```

v1v: Gotcha, thanks for the clarification

nxei:
> What does Gha stand for?

Gha stands for github actions.

nxei:
>Please read the previous comment regarding the pre-commit, I have not seen any new changes to fix the linted error.

I installed pre-commit and used it to fix problems with spaces.

nxei:
> Can you please fix the commit linting defect? Those defects can be detected locally with the pre-commit, the installation details can be found in https://github.com/elastic/apm-pipeline-library/blob/master/docs/PRECOMMIT.md#installation

>Then every time there is a git commit the pre-commit will evaluate the existing rules and will fail if a defect is detected.

>The good thing is some of the linting rules do fix those defects so, you only need to git add and git commit again.

Installed pre-commit.

 >Do you think a new Unit test case could help to document this requirement? If so, please add one.

I agree that unit tests needed and I created 3 unit tests.

v1v (approved):

PR #1415, opened 2021-11-25
Deleted setting runner parameter as mandatory argument for workflow.

nxei:
## What does this PR do?

Deleted setting runner parameter as mandatory argument for workflow.

## Why is it important?

This parameter is specific to the workflow so it should not be set by default and added to any workflows mandatory.

### DIFF
--- a/vars/githubWorkflowRun.groovy
+++ b/vars/githubWorkflowRun.groovy
@@ -26,8 +26,8 @@
            workflow: "build.yml",
            ref: "main",
            parameters: [
-             path: "filebeat"
-           ],
+             path: "filebeat",
+             runner: "ubuntu-latest"],
            credentialsId: "github-workflow-token",
            ghVersion: "2.1.0"]
        def runId = githubWorkflowRun.triggerGithubActionsWorkflow(args)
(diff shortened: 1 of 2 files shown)


v1v (changes requested):
I agree that probably a default value should not be defined in favour of validating if `args.containsKey('runner')` returns true and if so then add the runner to the `inputs` map.


Why do I say so? 

As far as I see the consumer https://github.com/elastic/beats/pull/29148/files#diff-3569b5feda562f1b9cb8c965c02dc80df7a1969e39e0a8ecb0439553a248222aR16-R19 uses an input parameter


Therefore, either the consumer should not have defined the runner, or this step should allow the customisation.

nxei:
> As far as I see the consumer https://github.com/elastic/beats/pull/29148/files#diff-> 3569b5feda562f1b9cb8c965c02dc80df7a1969e39e0a8ecb0439553a248222aR16-R19 uses an input parameter

> Therefore, either the consumer should not have defined the runner, or this step should allow the customisation. 

https://github.com/elastic/beats/blob/2d04883dcaef86723cc7183e7f8d4aa9ea2eba58/Jenkinsfile#L364-L369

The runner parameter is specific to thew macos-bild.yml workflow. It can be specified as entry in parameters map, like this:

```
         parameters: [
            runner: 'ubuntu-18.04',
            pr: env.CHANGE_ID,
            path: args.directory,
            command: args.command,
            go_version: env.GO_VERSION,
          ])
```

It will allow specify required parameters and skip not needed so customization is  possible.

v1v (approved):
Thanks for the clarification

PR #904, opened 2021-01-14
[chore]: delegate delete to the core

v1v:
See https://issues.jenkins.io/browse/JENKINS-41805 or https://github.com/jenkinsci/jenkins/pull/4824

## What does this PR do?

Avoid the explicit of deleting the workspace by using the default core behaviour

## Why is it important?

Ensure the workspace is left and Jenkins can track that also @lib and @scripts and other siblings folder can be deleted.

I assume by deleting the workspace those folders won't be removed though

## Related issues
Closes #ISSUE

PR #1416, opened 2021-11-25
ci: delete workspace after notifying

v1v:
## What does this PR do?

Reenable the deleteDir in the notifyBuildResult.

## Why

I saw some weird behaviour with the test reporting in the GitHub comments.

A file timestamp was wrong 

image

Tests are not present

image

But the json with the tests files is filled with probably some data from a previous build:

image

I wonder if something is cached in the Jenkins controller ?

### Issues

Initially enabled in https://github.com/elastic/apm-pipeline-library/pull/511

Removed in https://github.com/elastic/apm-pipeline-library/pull/904

### DIFF
--- a/vars/notifyBuildResult.groovy
+++ b/vars/notifyBuildResult.groovy
@@ -125,6 +125,10 @@ def call(Map args = [:]) {
           }
         }
       }
+      // Ensure we don't leave any leftovers if running in the jenkins controller.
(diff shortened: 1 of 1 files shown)


cachedout (approved):
Interesting. Nice catch.

kuisathaverat (approved):

PR #1417, opened 2021-11-29
Remove support for index jenkins-build

v1v:
## What does this PR do?

Remove the support for the deprecated `jenkins-build` index.

## Why is it important?

This index does contain all the executed tests and in some cases those tests can be a few thousands which could cause some issues in the controller when accessing the BlueOcean RestAPI.

image

In addition, we already moved to some normalised data structure with some other indexes long time ago (ci-builds, ci-tests).

## Actions

- [x] Confirm this particular index is not anymore consumed.
- [x] Update the docs

## Issues

I discovered in https://github.com/elastic/apm-agent-nodejs/pull/2471 that 200.000 reported tests might cause issues when fetching those tests in order to send them back to our ES cluster.


v1v:
@cachedout , I need your input here, no rush, but IIRC, all the flakiness stuff is not anymore consuming the deprecated index but the `ci-tests` index?

kuisathaverat (approved):

v1v:
I just verified we don't use this particular index

image

v1v:
There are some other tools that consume this index, but we can change it to use the `ci-builds` or `ci-tests` indexes instead. I think they are CLI tools, so we should be safe to merge this PR IMO

mdelapenya (approved):
LGTM, although I have one question about something I do not understand:

We are removing the limit filter in the shell script (https://github.com/elastic/apm-pipeline-library/pull/1417/files#diff-27f285536c55231dd2c04cfbafa1df0a2ccd01a8687258180fc9521469f26dcfR493) but adding it to all the modified JSON files

Could you please clarify? 🙏

cachedout:
> all the flakiness stuff is not anymore consuming the deprecated index but the ci-tests index?

Correct. The flaky test analyzer [consumes](https://github.com/elastic/observability-dev/blob/main/apps/automation/flaky-test-analyzer/lib/__init__.py#L28) the `ci-test` index.

v1v:
> We are removing the limit filter in the shell script (https://github.com/elastic/apm-pipeline-library/pull/1417/files#diff-27f285536c55231dd2c04cfbafa1df0a2ccd01a8687258180fc9521469f26dcfR493) but adding it to all the modified JSON files

I don't quite understand the question.

The BO entrypoint for the `/tests/?limit=10000000"` has been replaced for the `tests/?status=FAILED&limit=1000`, this should avoid issues with massive tests results.

L493 consumes the file generated in L480. L480 is the one using `tests/?status=FAILED&limit=1000`

I don't know if I was able to answer the question

cachedout:
@v1v Regarding these changes:

Previously, we had a `test` key in the `jenkins-build` index which contained a complete list of tests which ran for a given build and the result of each.

I understand that we now have individual test _failures_ in the `ci-tests*` indices but we still don't have a complete list of tests which were executed for any given build.  

If I understand this new schema correctly, it is no longer possible to retrieve a a complete list of tests which were executed for a given build by querying the data we're keeping in Elasticsearch. Is that correct?

v1v:
> If I understand this new schema correctly, it is no longer possible to retrieve a a complete list of tests which were executed for a given build by querying the data we're keeping in Elasticsearch. Is that correct?

That's correct

`jenkins-build` contained a bunch of data, but it didn't scale well when the total number of tests were in the magnitude of 10.000, the bulk update in ES used to take several minutes.

`ci-tests` only contains the failed tests as you mentioned.

Potentially, https://github.com/mdelapenya/junit-otel-reporter-plugin/ could be the way to go to digest all the tests as spans, but we are still in an early stage.

On the other hand, we could also use a batch process to digest all the executed builds and send the UT data in async way

cachedout:
> That's correct

Thanks for the response, @v1v 

This is a bit of a shame since the [Integrations Test Reporter](https://github.com/elastic/observability-dev/tree/main/apps/automation/integrations/reporter) depends on being able to fetch a full list of tests which were executed. Without a list of executed tests, we have no way of knowing whether or not a package is being tested or not.

To work around this, I'm afraid that we'll need to rewrite the report generator to [query Jenkins directly](https://beats-ci.elastic.co/job/Ingest-manager/job/integrations/job/master/1316/artifact/build/test-results/).

PR #1429, opened 2021-11-30
Add download step

v1v:
## What does this PR do?

Add new step to download either using wget or curl.

Refactor the consumers

## Why is it important?

Keep our code simpler

### DIFF
--- a/vars/download.groovy
+++ b/vars/download.groovy
@@ -1,0 +1,30 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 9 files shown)


vars/download.groovy
```diff
@@ -0,0 +1,30 @@
...
+/**
+ Download the given URL regardless of the tool.
+
+ download(url: 'https://....', output: 'gsutil.tar.gz')
+
+*/
+def call(Map args = [:]) {
+  def url = args.containsKey('url') ? args.url : error('download: url parameter is required')
+  def output = args.containsKey('output') ? args.output : error('download: output parameter is required')
+  if (!downloadWithWget(url: url, output: output)) {
```
mdelapenya, line 27: This will try wget first, right? Is it possible for a consumer to select `curl` as first choice? Does it make sense this use case?

v1v: This is just a refactor, so if needed we can change behaviour but I'd say to do it in a follow up. What's your use case?

mdelapenya: I was thinking in a consumer using: download(tool: 'curl'...), but I've just realised that I could simply do 'downloadWithCurl' 😄

mdelapenya (approved):

kuisathaverat (approved):

PR #1435, opened 2021-12-02
Allow to customise the vault token

v1v:
## What does this PR do?

Allow to customise the vault arguments

## Why is it important?

When running in a different CI controller those credentials might differ


vars/withAWSEnv.groovy
```diff
@@ -37,7 +37,7 @@ def call(Map args = [:], Closure body) {
     if (forceInstallation || !isInstalled(tool: 'aws', flag: '--version', version: version)) {
       downloadAndInstall(awsUtilLocation, version)
     }
-    def props = getVaultSecret(secret: secret)
+    def props = getVaultSecret(args)
```
v1v, line 40: We delegate the arguments validation in `getVaultSecret` therefore we can pass the args map directly

cachedout: Is this every going to be untrusted input? For example, are we setting ourselves up for a situation where a user could provide a parameter to a job and retrieve an arbitrary vault secret? I might feel a bit better if we just took a string as an argument and then switched to one of a couple of known strings for vault paths. WDYT?

v1v: > Is this every going to be untrusted input?  

In theory, non-elasticians cannot run anything in the CI. So, that's the current paradigm we are following.

>  are we setting ourselves up for a situation where a user could provide a parameter to a job and retrieve an arbitrary vault secret? 

They can nowadays use the `getVaultSecret`

> WDYT?

We could follow that approach, but `getVaultSecret` already provide those parameters anyway. I didn't do much but allowing that customisation to the `withAWSEnv`.

Do you think we should raise an Issue to describe this and what we could do? Or do you prefer to stop here and think about the implications a bit more?  As a mention, people can use `getVaultSecret` with the secret_id and role_id. Though, those IDs are credentials, so they are masked too

cachedout: > They can nowadays use the getVaultSecret

Right, I know. I'm going to move this discussion to a private repo and I'll link back here.

kuisathaverat (approved):

cachedout (approved):

PR #1437, opened 2021-12-02
ci: run a few stages in parallel to speed up

v1v:
## What does this PR do?

Run in parallel a few stages to speed up the CI feedback

## Why is it important?

Windows Google storage takes a few minutes:

image

## Issue

Maybe https://github.com/elastic/apm-pipeline-library/pull/1438 could be the alternative?


v1v:
I'd say we can move away from this approach for now, and once we use k8s we can then use it

PR #1458, opened 2021-12-15
[jjbb][main] neutral naming for observability-robots

v1v:
## What does this PR do?

Rename master to main

## Why is it important?

Neutral naming of git branches


.ci/buildDockerImages.groovy
```diff
@@ -563,6 +566,7 @@ def dockerLoginElasticRegistry(){
 
 def buildDockerImage(args){
   String repo = args.containsKey('repo') ? args.repo : error("Repository not valid")
+  String branch = args.get('branch', 'master')
```
cachedout, line 569:
```suggestion
  String branch = args.get('branch', 'main')
```

Not `main`?

v1v: no yet, by default, others are not yet migrated, so we cannot apply that default value. I added a new signature then I can explicitly opt-in when rolling out the main branch.

My plan, is to at the very end to change the default to `main`

cachedout (approved):

kuisathaverat (approved):

PR #1462, opened 2021-12-16
fix: improve attributes access

kuisathaverat:
## What does this PR do?

* improve the access to object attributes

## Why is it important?

In some cases, some of the attributes cannot be defined and this causes errors like the following.

```
[2021-12-16T03:42:28.564Z] >               xfailed = item._store.get(_pytest.skipping.xfailed_key, None)
[2021-12-16T03:42:28.564Z] E               AttributeError: module '_pytest.skipping' has no attribute 'xfailed_key'
```


resources/scripts/pytest_otel/src/pytest_otel/__init__.py
```diff
@@ -235,32 +236,41 @@ def pytest_runtest_call(item):
...
-                span.set_attribute("test.error", "{}".format(sys.last_value.args[0]))
-            except IndexError:
-                span.set_attribute("test.error", "{}".format("unknown"))
-            span.set_attribute("test.last_value", "{}".format(sys.last_value))
-            span.set_attribute("test.last_type", "{}".format(sys.last_type))
-
-            xfailed = item._store.get(_pytest.skipping.xfailed_key, None)
-            if xfailed:
-                span.set_attribute("test.xfailed", "{}".format(xfailed.reason))
+            if hasattr(last_value, "args") and len(getattr(last_value, 'args')) > 0:
```
cachedout, line 259:
```suggestion
            if hasattr(last_value, "args") and len(getattr(last_value, 'args', [])) > 0:
```
If `args` is undef, then `getattr` will return None and `len()` will cause a `TypeError`. Setting the default to an empty list should prevent this.

resources/scripts/pytest_otel/setup.cfg
```diff
@@ -5,7 +5,7 @@ long_description = file: README.md
 long_description_content_type = text/markdown
 url = https://github.com/elastic/apm-pipeline-library/tree/master/resources/scripts/pytest_otel
 maintainer = Ivan Fernandez Calvo
-version = 0.0.4
+version = 0.0.5
```
kuisathaverat, line 8:
```suggestion
version = 0.0.6
```

v1v (approved):

cachedout (approved):

PR #1468, opened 2021-12-17
Update main

v1v:
~~This should be merged beforehand then we can rename. Otherwise we are blocked by the JJBB automation~~ A read-only master branch is in place, so we should not have any issues


.ci/beats-webhook.groovy
```diff
@@ -51,7 +51,7 @@ pipeline {
      printPostContent: false,
      silentResponse: true,
      regexpFilterText: '$ref-$x_github_event',
-     regexpFilterExpression: '^(refs/tags/current|refs/heads/master/.+)-comment$'
+     regexpFilterExpression: '^(refs/tags/current|refs/heads/master/.+|refs/heads/main/.+)-comment$'
```
cachedout, line 54: Are we ready to change this?

v1v: It's not replacing master but adding it. The reason was to avoid forgetting this change in the future, so I decided to add it to the regex then it should be harmless

cachedout (approved):

PR #1471, opened 2021-12-20
docker: no cache flag when building the image

v1v:
## What does this PR do?

Use `--no-cache                Do not use cache when building the image`

## Why is it important?

Node:16 and Node:17 cached docker images point to node:8

## Related issues

Relates to https://github.com/elastic/ecs-logging-nodejs/issues/112


cachedout (approved):
Good discovery! 🏅

PR #1472, opened 2021-12-20
PoC for running a job in another controller

v1v:
## What does this PR do?

Similar to the `build` step but using a new step to interact with a different CI controller.

## Why is it important?

Provide a helper function that validates Jobs and CI Controllers before running the job in a different controller.


vars/buildInController.groovy
```diff
@@ -0,0 +1,38 @@
...
+def call(Map args = [:]){
+  def job = args.job
+  def parameters = args.parameters
+  def wait = args.get('wait', true)
+  def propagate = args.get('propagate', true)
+  def quietPeriod = args.get('quietPeriod', 1)
+  def controller = args.containsKey('controller') ? args.controller : error('')
+
+  // Do we want to validate the Job is in the approval list of jobs?
+  // Do we want to validate the controller is in the approval list of controllers?
```
mtojek, line 35: As I commented in the other issue, do you know if there is any RBAC-like feature in Jenkins, which could replace this security mechanism?

v1v: > As I commented in the other issue, do you know if there is any RBAC-like feature in Jenkins, which could replace this security mechanism?

There is, it's up to the system owner to provide that particular level of granularity but it requires a service account if we use that option on the other hand

vars/buildInController.groovy
```diff
@@ -0,0 +1,38 @@
...
+  buildInController(job: 'foo',
+                    parameters: [string(name: "my param", value: some_value)]
+                    controller: 'my-new-controller')
+*/
+def call(Map args = [:]){
+  def job = args.job
+  def parameters = args.parameters
+  def wait = args.get('wait', true)
+  def propagate = args.get('propagate', true)
+  def quietPeriod = args.get('quietPeriod', 1)
```
mdelapenya, line 31: I guess these three args should be documented in the txt file too

vars/buildTokenTriggerExt.groovy
```diff
@@ -0,0 +1,42 @@
...
+def call(Map args = [:]){
+  def wait = args.get('wait', false)
+  def propagate = args.get('propagate', false)
+  String credentialsId = args.get('credentialsId', '')
+  if (!isPluginInstalled(pluginName: 'build-token-trigger')) {
+    error('buildTokenTriggerExt: build-token-trigger plugin is not available')
+  }
+
+  if (wait || propagate) {
+    echo 'TBD'
```
v1v, line 38: This is now the tricky part :)

vars/buildTokenTriggerExt.groovy
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+def waitFor(Map args = [:]) {
+  // for testing purposes, let's mock an error
+  if (args.get('job', '').equals('mock-test-error-for-unit-testing')) {
+    error("waitFor: mock an error for the UTs.")
+  }
+  def to_url = args.jenkinsUrl + "/job/" + args.job
+  return processedJob(httpRequest(url: to_url))
```
v1v, line 74: What job to query and what credentials to be consumed.

vars/buildTokenTriggerExt.groovy
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// TODO: to be implemented
+def processedJob(request) {
+  // Data structure:
+  //  - finished if the job is still running
+  //  - failed if the job finished and failed
+  //  - status, SUCCESS, FAILURE, UNSTABLE, ABORTED, NOT_BUILT
+  return [ finished: false, failed: false, status: 'NOT_BUILT' ]
```
v1v, line 83: Once we know the data structure from the response in https://github.com/elastic/apm-pipeline-library/pull/1472/files#diff-6b4326b3ad4ab476d4fe520474c1ec008c36265d310b799873d93b0741985f3eR74 then we can do the transformation

v1v:
Most likely this implementation is obsoleted in favour of https://plugins.jenkins.io/Parameterized-Remote-Trigger/

mtojek:
> Most likely this implementation is obsoleted in favour of https://plugins.jenkins.io/Parameterized-Remote-Trigger/

Maybe we should rediscuss the plugin idea vs. this solution with @jonahbull ?

PR #1476, opened 2021-12-20
Beats: notify 7 next minor branch and 7 current branch

v1v:
## What does this PR do?

Support two 7 minor branches (7.16 and 7.17)

## Why is it important?
`7.17` was not reported while `7.16` is also active.

Thanks @ruflin for pointing out about this

## Follow ups

How to dynamically know if there are two active 7.minor branches

## Issues

Requires https://github.com/elastic/apm-pipeline-library/pull/1470


ruflin (approved):
Hopefully this can be cleaned up in the near future again to only have 1 active branch.

PR #1495, opened 2022-01-12
Allow to customise the vault token in withGCPEnv

v1v:
## What does this PR do?

Enable vault customisation for the withGCPEnv

## Why is it important?

Similar to https://github.com/elastic/apm-pipeline-library/pull/1435


v1v:
Thanks @mtojek for pointing to https://github.com/elastic/apm-pipeline-library/blob/ed81872efa5c15f1e4d8721ae732f7f96d5c1aea/vars/getVaultSecret.groovy#L28-L29 therefore we don't need anything this change since implicitlye the library detects whether it should use a different vault configuration when running in a different CI controller :)

mtojek (approved):

PR #1496, opened 2022-01-13
feat: update the attribute references

kuisathaverat:
## What does this PR do?

* update the attribute references
* * prepare for send metrics

## Why is it important?

To use the ECS for the attributes is key to having the same experience across test frameworks. Metrics also are a useful data to collect.


resources/scripts/pytest_otel/src/pytest_otel/__init__.py
```diff
@@ -18,6 +19,12 @@
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 from opentelemetry.trace.status import Status, StatusCode
 
+# from opentelemetry import metrics
```
kuisathaverat, line 22: the metrics are available on >1.7.0, but we cannot update yet due to that version has a bug on the headers, this bug breaks the authentication with headers

mdelapenya (approved):

v1v (approved):

PR #1497, opened 2022-01-13
withGCPEnv: support for value default field in the vault secret

v1v:
## What does this PR do?

Support the `value` field in addition to the existing support for the `credentials` one.

## Why is it important?

Some secrets are managed with terraform and therefore the field is not matching the default one.

Let's keep backward compatibility with the existing credentials one.

For context, `credentials` was aimed to be the mandatory field in order to early detect wrong vault secrets, since `value` is the default field then a secret could be used wrongly and the failure will not be detected earlier but when running the `gcloudAuth` method.

### DIFF
--- a/vars/withGCPEnv.groovy
+++ b/vars/withGCPEnv.groovy
@@ -46,10 +46,7 @@ def call(Map args = [:], Closure body) {
         error "withGCPEnv: Unable to get credentials from the vault: ${props.errors.toString()}"
       }
       def value = props?.data
-      def credentialsContent = value?.credentials
-      if (!credentialsContent?.trim()) {
-        error "withGCPEnv: Unable to read the credentials value"
-      }
+      def credentialsContent = readCredentialsContent(value)
       writeFile(file: secretFileLocation, text: credentialsContent)
       gcloudAuth(secretFileLocation)
     } else {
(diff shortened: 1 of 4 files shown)


vars/withGCPEnv.groovy
```diff
@@ -81,6 +78,23 @@ def call(Map args = [:], Closure body) {
...
 
+/**
+* Read the vault secret and look for the required fields.
+* * credentials field is the one initially supported and kept for backward compatibility reasons.
+* * value field is the fallback field, this is the case used
+*/
+def readCredentialsContent(Map vaultSecretContent) {
+  def credentialsContent = vaultSecretContent?.credentials
+  if (credentialsContent?.trim()) {
+    return credentialsContent
```
mtojek, line 89: nit: out of curiosity I would drop a logger here to see which field is used :)

mtojek (approved):

kuisathaverat (approved):

vars/withGCPEnv.groovy
v1v: See https://github.com/elastic/apm-pipeline-library/pull/1497/commits/121e3c3400d10f0c650f3de20cc7cbe4077cb798

PR #1503, opened 2022-01-19
Minor index

mdelapenya:
- chore: copy stack version bumps from oblt-dev
- chore: use apm-pipeline library as repository for stack version bumps
- feat: add a step to retrieve the branch name supporting multiple minors
- chore: use new step to retrieve the branch name for the stack update

## What does this PR do?
It adds a new step that supports passing a branch with a specific format:

- major.minor
- major.<minor>
- major.<minor-$index>

The index will be used to retrieve a previous minor version of the stack

## Why is it important?
It could be the case where there are multiple minors in development at the same time, 7.16 and 7.17, and with previous behaviour, the automation discarded those cases where the project needed a bump in two minors (See e2e-tests, that needs bumps in 7.16 and 7.17)

### DIFF
--- a/.ci/.bump-stack-version.yml
+++ b/.ci/.bump-stack-version.yml
@@ -40,6 +40,7 @@ projects:
       - main
       - 8.<minor>
       - 7.<minor>
+      - 7.<minor-1>
     enabled: true
     reusePullRequest: false
     labels: dependency,backport-skip
--- a/.ci/.bump-stack-version.yml
+++ b/.ci/.bump-stack-version.yml
@@ -1,0 +1,55 @@
+---
+## This is the file that contains what projects should bump their elastic stack
+## versions.
+##
+## The data structure:
+##  repo: the repository name
+##  script: the relative path to the script in charge to bump the version
+##  branches: the list of branches that should benefit from this automation.
+##  enabled: whether the automation has been enabled for this project.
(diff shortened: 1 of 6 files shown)


.ci/.bump-stack-version.yml
```diff
@@ -0,0 +1,55 @@
...
+      - "6.8"
+    enabled: true
+    reusePullRequest: false
+    labels: dependency,build-monitoring,backport-skip,Team:Beats-On-Call
+  - repo: e2e-testing
+    script: .ci/bump-stack-version.sh
+    branches:
+      - main
+      - 8.<minor>
+      - 7.<minor>
```
v1v, line 42: do you want to use minor-1?

v1v (approved):

.ci/.bump-stack-version.yml
mdelapenya: Yes, let me change it, and I need another PR removing the file from oblt-dev

v1v: Since it uses `main` -> https://github.com/elastic/apm-pipeline-library/pull/1503/files#diff-eba3c46469d30c72b2ce74f5be13c5b7d85e69e0d8ef95dba6e5bf8e3f09a9e8R20 as long it gets merged then it should work :)

PR #1519, opened 2022-01-31
github commands: enable windows support for the java-agent

v1v:
## What does this PR do?

Enable new GitHub command support for the APM Agent Java.

## Why is it important?

Caused by https://github.com/elastic/apm-agent-java/pull/1705


SylvainJuge (approved):
LGTM.

Do you think there could be a way to directly define those once in each agent repo ? For example directly into the pipeline or close to where the trigger regex is being defined ?

v1v:
> Do you think there could be a way to directly define those once in each agent repo ? For example directly into the pipeline or close to where the trigger regex is being defined ?

I've not found a way to define them and add some description easily. Maybe as you said, a file in the repository could be the one to be read and sent over, we can iterate on this idea, if the file is defined then it's read, though it makes the step to rely on the workspace to be able to read the file.

mdelapenya (approved):

PR #1533, opened 2022-02-08
feat: move stackVersions to use bumpUtils

kuisathaverat:
## What does this PR do?

* bump patch version to 8.1
* change stackVersions to use 

## Why is it important?

It reduces the manual updates we have to make

### DIFF
--- a/vars/stackVersions.groovy
+++ b/vars/stackVersions.groovy
@@ -44,22 +44,15 @@ def version(value, args = [:]){
 }
 
 def edge(Map args = [:]){
-  // TODO: this should not use a hardcoded string but bmptUtils.getNextMinorReleaseFor8()
-  return version("8.1.0", args)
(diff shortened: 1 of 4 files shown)


v1v (approved):

vars/stackVersions.groovy
```diff
@@ -44,22 +44,15 @@ def version(value, args = [:]){
...
-  return version("8.1.0", args)
+  return version(bumpUtils.getNextMinorReleaseFor8(), args)
 }
 
-def dev(Map args = [:]){
-  // IMPORTANT: this variable is needed to automate the elastic stack bump
-  def devVersion = "7.17.1"
-  // TODO: this should not use a hardcoded string but bmptUtils.getNextMinorReleaseFor7()
-  return version(devVersion, args)
+def dev(Map args = [:]){)
```
kuisathaverat, line 50:
```suggestion
def dev(Map args = [:]){
```

Issue #1449, opened 2021-12-09
Enable notifyBuildResult to run in a different CI Controller


v1v:
### What

`notifyBuildResult` to prepare the correct credentials depending on in which CI controller runs

### Why 

@mtojek discovered that the `internal` controller does not have access to the same credentials therefore the `notifyBuildResult` cannot be used to notify the status as long as it uses the default parameters.

v1v:
```
[](https://internal-ci.elastic.co/blue/organizations/jenkins/package_storage%2Fpackage-storage-infra-mbp/detail/PR-34/1/pipeline#step-289-log-2)Unable to find project for artifact copy: package_storage/package-storage-infra-mbp/PR-34

This may be due to incorrect project name or permission settings; see help for project name in job configuration.
```

PR #1537, opened 2022-02-10
Support internal controller with different permissions

v1v:
## What does this PR do?

Add support for the internal controller

## Why is it important?

Copy Artifacts between jobs is not supported in some particular cases

## Related issues
Closes https://github.com/elastic/apm-pipeline-library/issues/1449

### DIFF
--- a/vars/githubPrComment.groovy
+++ b/vars/githubPrComment.groovy
@@ -108,7 +108,12 @@ def editComment(id, details) {
 
 def getCommentFromFile(Map args = [:]) {
   def commentFile = args.commentFile
-  copyArtifacts(filter: commentFile, flatten: true, optional: true, projectName: env.JOB_NAME, selector: lastWithArtifacts())
+  try {
+    copyArtifacts(filter: commentFile, flatten: true, optional: true, projectName: env.JOB_NAME, selector: lastWithArtifacts())
+  } catch(e) {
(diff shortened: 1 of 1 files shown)


vars/githubPrComment.groovy
```diff
@@ -108,7 +108,12 @@ def editComment(id, details) {
 
 def getCommentFromFile(Map args = [:]) {
   def commentFile = args.commentFile
-  copyArtifacts(filter: commentFile, flatten: true, optional: true, projectName: env.JOB_NAME, selector: lastWithArtifacts())
+  try {
+    copyArtifacts(filter: commentFile, flatten: true, optional: true, projectName: env.JOB_NAME, selector: lastWithArtifacts())
+  } catch(e) {
+    // Some CI controllers don't grant access to copy artifacts between jobs.
```
mdelapenya, line 114: Could we print/wrn the error log to understand the causes?

v1v: Sure

vars/githubPrComment.groovy
```diff
@@ -108,7 +108,12 @@ def editComment(id, details) {
...
+    return ''
```
v1v, line 115:
```suggestion
    log(level: 'INFO', text: "Could not copy artifacts from the build. ${err.toString()}")
    return ''
```

kuisathaverat (approved):

PR #1550, opened 2022-02-15
support push docker image step

v1v:
## What does this PR do?

Refactor a common step used in Beats, elastic-agent and fleet-server that could benefit us to reduce the complexity in those pipelines with something that's tested accordingly.

You can see how it works in:
- https://github.com/elastic/beats/pull/30414

### DIFF
--- a/vars/README.md
+++ b/vars/README.md
@@ -184,6 +184,7 @@ Override the `build` step to highlight in BO the URL to the downstream job.
 
 ```
 build(job: 'foo', parameters: [string(name: "my param", value: some_value)])
+build 'foo'
 ```
 
 See https://jenkins.io/doc/pipeline/steps/pipeline-build-step/#build-build-a-job
@@ -194,9 +195,11 @@ Builds the Docker image for Kibana, from a branch or a pull Request.
 ```
 buildKibanaDockerImage(refspec: 'main')
 buildKibanaDockerImage(refspec: 'PR/12345')
+buildKibanaDockerImage(refspec: 'cf25ac3d1f8edff8f20003add707bfdc85d89fff', depth: 10)
+
 ```
 
-* refspec: A branch (i.e. main), or a pull request identified by the "pr/" prefix and the pull request ID.
+* refspec: A branch (i.e. main), a commit SHA, a tag, or a pull request identified by the "pr/" prefix and the pull request ID.
 * packageJSON: Full name of the package.json file. Defaults to 'package.json'
 * baseDir: Directory where to clone the Kibana repository. Defaults to "${env.BASE_DIR}/build"
 * credentialsId: Credentials used access Github repositories.
@@ -205,6 +208,9 @@ buildKibanaDockerImage(refspec: 'PR/12345')
 * dockerRegistrySecret: Name of the Vault secret with the credentials for logining into the registry. Defaults to 'secret/observability-team/ci/docker-registry/prod'
 * dockerImageSource: Name of the source Docker image when tagging. Defaults to '${dockerRegistry}/kibana/kibana'
 * dockerImageTarget: Name of the target Docker image to be tagged. Defaults to '${dockerRegistry}/observability-ci/kibana'
+* reference: Path to the Git reference repo to improve checkout speed. Default to '/var/lib/jenkins/kibana.git'
+* depth: Number of commits pull down in the Git shallow clone. Default to 1
+* shallow: Enable shallow cloning. Default to true.
 
 ## buildStatus
 Fetch the current build status for a given job
@@ -667,7 +673,7 @@ update the name of the branch when a new minor release branch is created.
 
 ```
 // Return the branch name for the main, 8.minor and 8.next-minor branches
-def branches = getBranchesFromAliases(aliases: ['main', '8.<minor>', '8.<minor-minor>'])
+def branches = getBranchesFromAliases(aliases: ['main', '8.<minor>', '8.<next-minor>'])
 
 ```
 
@@ -1543,6 +1549,7 @@ Upload the given pattern files to the given bucket.
 * credentialsId: The credentials to access the repo (repo permissions). Optional. Default to `JOB_GCS_CREDENTIALS`
 * pattern: The file to pattern to search and copy. Mandatory.
 * sharedPublicly: Whether to shared those objects publicly. Optional. Default false.
+* extraFlags: Extra flags to use with gsutil cp. Optional
 
 ## gsutil
 Wrapper to interact with the gsutil command line. It returns the stdout output.
@@ -2526,6 +2533,66 @@ with the given headers.
 
 __NOTE__: It requires *Nix where to run it from.
 
+## pushDockerImages
+Publish docker images in the given docker registry. For such, it
+retags the existing docker images and publish them in the given
+docker namespace.
+
+It uses a map of images, this map contains an entry for each docker image
+to be pushed, what architecture and the name of the docker image to be pushed.
+
+The version is required to tag the docker image accordingly and also it uses
+the snapshot if needed.
+
+```
+  // Given the filebeat project, and its generated docker
+  // images for the 8.2.0-SNAPSHOT and 2 different variants
+  // then publish them to the observability-ci namespace. In addition
+  // tag them as default, arch=amd64 is the default tag image
+  pushDockerImages(
+    registry: "my-registry",
+    secret: "my-secret",
+    version: '8.2.0',
+    snapshot: true,
+    images: [
+      [ source: "beats/filebeat", arch: 'amd64', target: "observability-ci/filebeat"],
+      [ source: "beats/filebeat-ubi8", arch: 'amd64', target: "observability-ci/filebeat-ubi8"]
+    ]
+  )
+```
+
+```
+  // Given the filebeat project, and its generated docker
+  // images for the 8.2.0-SNAPSHOT and 2 different variants
+  // then publish them to observability-ci
+  // Source images follow the format:
+  //   - "my-registry/beats/filebeat:8.2.0-SNAPSHOT"
+  //   - "my-registry/beats-ci/filebeat-cloud:8.2.0-SNAPSHOT"
+  // Generated images follow the format:
+  //   - "my-registry/observability-ci/filebeat:8.2.0-SNAPSHOT"
+  //   - "my-registry/observability-ci/filebeat-cloud:8.2.0-SNAPSHOT"
+  //   - "my-registry/observability-ci/filebeat:8.2.0-SNAPSHOT-amd64"
+  //   - "my-registry/observability-ci/filebeat-cloud:8.2.0-SNAPSHOT-amd64"
+  pushDockerImages(
+    registry: "my-registry",
+    secret: "my-secret",
+    version: '8.2.0',
+    snapshot: true,
+    images: [
+      [ source: "beats/filebeat", arch: 'amd64', target: "observability-ci/filebeat"],
+      [ source: "beats-ci/filebeat-cloud", arch: 'amd64', target: "observability-ci/filebeat-ubi8"]
+    ]
+  )
+```
+
+* secret: the docker secret
+* registry: the docker registry
+* version: what version
+* snapshot: snapshot support
+* images: list of the docker image to be retagged to, architecture and docker image to be pushed to.
+
+__NOTE__: It requires *Nix where to run it from.
+
 ## randomNumber
 it generates a random number, by default the number is between 1 to 100.
 
@@ -3049,9 +3116,10 @@ not overridden once they are created.
 ```
 
 * repo: The GitHub repository name. Optional. Default to `REPO`
-* bucket: The Google Storage bucket name. Mandatory
+* bucket: The Google Storage bucket name. Optional. Default to `JOB_GCS_BUCKET`
 * credentialsId: The credentials to access the repo (repo permissions). Optional. Default to `JOB_GCS_CREDENTIALS`
 * pattern: The file to pattern to search and copy. Optional. Default to `build/distributions/**/*`
+* folder: The folder to be added to the calculated bucket uri folder. Optional.
 
 NOTE: It works with the Multibranch Pipeline only, therefore it requires to use `gitCheckout` to be able to populate the
       gitBaseCommit.
--- a/vars/README.md
+++ b/vars/README.md
@@ -3034,6 +3034,28 @@ updateGithubCommitStatus(message: 'Build result.')
 
 It requires [Github plugin](https://plugins.jenkins.io/github)
 
+## uploadPackagesToGoogleBucket
+Upload the given pattern files to the given bucket using an opinionated folder structure:
+
+* <repo>/snapshots             -> if a branch.
+* <repo>/pull-requests/pr-<id> -> if a Pull Request.
+* <repo>/commits/<git-commit>  -> regardless of the type of a build.
+
+Snapshots and pr-<id> folders might contain files that are overridden while `commits/<git-commit>` are
+not overridden once they are created.
+
+```
+  uploadPackagesToGoogleBucket(pattern: 'file.txt', bucket: 'bucket', credentialsId: 'foo', repo: 'foo')
+```
+
+* repo: The GitHub repository name. Optional. Default to `REPO`
+* bucket: The Google Storage bucket name. Mandatory
+* credentialsId: The credentials to access the repo (repo permissions). Optional. Default to `JOB_GCS_CREDENTIALS`
+* pattern: The file to pattern to search and copy. Optional. Default to `build/distributions/**/*`
+
+NOTE: It works with the Multibranch Pipeline only, therefore it requires to use `gitCheckout` to be able to populate the
+      gitBaseCommit.
+
 ## whenFalse
 This step replaces those small scripts step blocks to check some condition,
 it simplifies Declarative syntax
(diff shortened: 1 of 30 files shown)


kuisathaverat (approved):

vars/pushDockerImages.txt
```diff
@@ -0,0 +1,38 @@
...
+```
+
+* secret: the docker secret
+* registry: the docker registry
+* arch: the supported arch (amd64 or arm64)
+* version: what version
+* snapshot: snapshot support
+* imageName: the docker image name.
+* variants: variants and docker namespace to copy from.
+* targetNamespace: what docker namespace to publish to.
```
kuisathaverat, line 36: `arch` and `variant` are complicated, it is not obvious what will be the tag, maybe it worth a few examples to understand how it works

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,139 @@
...
+def doTagAndPush(Map args = [:]) {
+  def name = args.imageName
+  def variant = args.variant
+  def sourceTag = args.sourceTag
+  def targetTag = args.targetTag
+  def sourceNamespace = args.sourceNamespace
+  def targetNamespace = args.targetNamespace
+  def registry = args.registry
+
+  def sourceName = "${registry}/${sourceNamespace}/${name}${variant}:${sourceTag}"
```
kuisathaverat, line 103: Not sure is correct

```
def registry = 'docker.elastic.co'
def sourceNamespace = 'observability-ci'
def name = 'metricbeat'
variant = 'cloud'
sourceTag = '8.0.0-SNAPSHOT'

docker.elastic.co/observability-ci/metricbeatcloud:8.0.0-SNPASHOT
```

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,139 @@
...
+  def targetName = "${registry}/${targetNamespace}/${name}${variant}:${targetTag}"
```
kuisathaverat, line 104: not sure is correct
```
def registry = 'docker.elastic.co'
def sourceNamespace = 'observability-ci'
def name = 'metricbeat'
variant = 'cloud'
sourceTag = '8.0.0-SNAPSHOT'

docker.elastic.co/observability-ci/metricbeatcloud:8.0.0-SNPASHOT
```

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,139 @@
...
+  def snapshot = args.get('snapshot', true)
+
+  // Transform version in a snapshot.
+  def sourceTag = version
+  def aliasVersion = ""
+  if (snapshot) {
+    // remove third number in version
+    aliasVersion = version.substring(0, version.lastIndexOf(".")) + "-SNAPSHOT"
+    sourceTag += "-SNAPSHOT"
+  }
```
kuisathaverat, line 64: I think that this logic should not be here, it is related to process aliases, but not related to pushes Docker images, also, it is not needed if you pass the correct tag.

```
def version = '8.0.0'

aliasVersion -> 8.0-SNAPSHOT
```

vars/pushDockerImages.groovy
v1v: I don't understand your question, `sourceNamespace` is never `observability-ci` but the one the consumer decides. That's exactly the same implementation as used to be -> https://github.com/elastic/beats/pull/30414/files#diff-5edb5ab93fc95960129eecb2ee3d8763a9c5e02b56e5b80126e0c5a52296ee8fL352

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,139 @@
...
+
+*/
+def call(Map args = [:]) {
+  if(!isUnix()){
+    error('publishToCDN: windows is not supported yet.')
+  }
+  def registry = args.containsKey('registry') ? args.registry : error('pushDockerImages: registry parameter is required')
+  def secret = args.containsKey('secret') ? args.secret : error('pushDockerImages: secret parameter is required')
+  def targetNamespace = args.containsKey('targetNamespace') ? args.targetNamespace : error('pushDockerImages: targetNamespace parameter is required')
+  def version = args.containsKey('version') ? args.version : error('pushDockerImages: version parameter is required')
```
kuisathaverat, line 51: It is stack-oriented variable, it limits the use of the step

vars/pushDockerImages.groovy
v1v: aha! `variant` should contain `-cloud` rather than `cloud`

vars/pushDockerImages.groovy
v1v: ditto https://github.com/elastic/beats/pull/30414/files#diff-5edb5ab93fc95960129eecb2ee3d8763a9c5e02b56e5b80126e0c5a52296ee8fL315-L321

vars/pushDockerImages.groovy
v1v: Bear in mind, this implementation did nothing but copy the existing one in the `packaging.groovy` file, so I didn't want to change the contract

vars/pushDockerImages.txt
```diff
@@ -0,0 +1,38 @@
...
+    registry: "my-registry",
+    arch: 'amd64',
+    version: '8.2.0',
+    snapshot: true,
+    imageName: 'filebeat',
+    variants: [
+      '' : 'beats',
+      'ubi8' : 'beats',
+      'cloud' : 'beats-ci',
+      'complete' : 'beats',
```
v1v, line 23:
```suggestion
      '' : 'beats',
      '-ubi8' : 'beats',
      '-cloud' : 'beats-ci',
      '-complete' : 'beats',
```

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,139 @@
...
+    registry: "my-registry",
+    arch: 'amd64',
+    version: '8.2.0',
+    snapshot: true,
+    imageName : 'filebeat',
+    variants: [
+      '' : 'beats',
+      'ubi8' : 'beats',
+      'cloud' : 'beats-ci',
+      'complete' : 'beats',
```
v1v, line 38:
```suggestion
      '-ubi8' : 'beats',
      '-cloud' : 'beats-ci',
      '-complete' : 'beats',
```

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,139 @@
...
+    ],
+    targetNamespace: 'observability-ci'
+  )
```
kuisathaverat, line 41: I think `version` and `arch` are redundant, the variants cover the need to push a matrix of Docker images. If we need to specialize the push we can do it in steps designed to implement that layer.

It is more straightforward to have in variants the source tag, the destination tag, and the namespace. even though you have to repeat some configuration values the code will be much simpler and easy to configure.

vars/pushDockerImages.txt
v1v: The above example is not enough?

vars/pushDockerImages.groovy
kuisathaverat: the values are examples to point you to the separator (`-`) issue

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,139 @@
...
+                      set -e
+                      echo "source: '${sourceName}' target: '${targetName}'"
+                      if docker image inspect "${sourceName}" &> /dev/null ; then
+                        docker tag "${sourceName}" "${targetName}"
+                        docker push "${targetName}"
+                      else
+                        echo "docker image ${sourceName} does not exist"
+                      fi""",
+                    returnStatus: true)
+    if (status > 0 && iterations < 3) {
```
mdelapenya, line 120: I'm trying to follow the values in the case the SH script fails, and I see a potential infinite loop here (or at least not sure if the `waitUntil` controls it, or nothing I can check in here https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#waituntil-wait-for-condition)

0. SH succeeded -> `iterations=1 && status=0` -> exits at the **else** conditional (true) ✅
1. SH fails -> `iterations=1 && status > 0` -> exits at the **first** conditional (false) ✅
2. SH fails -> `iterations=2 && status > 0` -> exits at the **first** conditional (false) ✅
3. SH fails -> `iterations=3 && status > 0` -> exits at the **second** conditional (false) 🚫
4. SH fails -> `iterations=4 && status > 0` -> exits at the **second** conditional (false) 🚫
5. SH fails -> `iterations=5 && status > 0` -> exits at the **second** conditional (false) 🚫

I'd propose this eval flow (may consider printing warning log in the else clause?):
0. SH succeeded -> `iterations=1 && status=0` -> exits at the **first** conditional (true) ✅
1. SH fails -> `iterations=1 && status > 0` -> exits at the **second** conditional (false) ✅
2. SH fails -> `iterations=2 && status > 0` -> exits at the **second** conditional (false) ✅
3. SH fails -> `iterations=3 && status > 0` -> exits at the **else** conditional (true). Desired at we reached the maximum ✅
4. SH fails -> `iterations=4 && status > 0` -> Not possible, as it's covered by above case ✅
5. SH fails -> `iterations=5 && status > 0` -> Not possible, as it's covered by above case ✅
6. 
```suggestion
    if (status == 0) {
        return true
    }
    else if (iteration < 3) {
        return false
    } else {
        return true
    }
```

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,139 @@
...
+      sleep 5 * iterations
+      return false
+    } else if (status > 0) {
+      return false
+    } else {
+      return true
+    }
```
mdelapenya, line 127: I'm trying to follow the values in the case the SH script fails, and I see a potential infinite loop here (or at least not sure if the `waitUntil` controls it, or nothing I can check in here https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#waituntil-wait-for-condition)

0. SH succeeded -> `iterations=1 && status=0` -> exits at the **else** conditional (true) ✅
1. SH fails -> `iterations=1 && status > 0` -> exits at the **first** conditional (false) ✅
2. SH fails -> `iterations=2 && status > 0` -> exits at the **first** conditional (false) ✅
3. SH fails -> `iterations=3 && status > 0` -> exits at the **second** conditional (false) 🚫
4. SH fails -> `iterations=4 && status > 0` -> exits at the **second** conditional (false) 🚫
5. SH fails -> `iterations=5 && status > 0` -> exits at the **second** conditional (false) 🚫

I'd propose this eval flow (may consider printing warning log in the else clause?):
0. SH succeeded -> `iterations=1 && status=0` -> exits at the **first** conditional (true) ✅
1. SH fails -> `iterations=1 && status > 0` -> exits at the **second** conditional (false) ✅
2. SH fails -> `iterations=2 && status > 0` -> exits at the **second** conditional (false) ✅
3. SH fails -> `iterations=3 && status > 0` -> exits at the **else** conditional (true). Desired at we reached the maximum ✅
4. SH fails -> `iterations=4 && status > 0` -> Not possible, as it's covered by above case ✅
5. SH fails -> `iterations=5 && status > 0` -> Not possible, as it's covered by above case ✅
6. 
```suggestion
    if (status == 0) {
        return true
    } else if (iteration < 3) {
        return false
    } else {
        return true
    }
```

mdelapenya: It could be reduced to:
```groovy
if (iteration < 3) {
        return false
} else {
        return true
}
```

mdelapenya: BTW, removing the sleep command because IIUC `waitUntil` implies it:
> Subsequent failures will slow down the delay between attempts up to a maximum of 15 seconds.

v1v:
> I think version and arch are redundant, the variants cover the need to push a matrix of Docker images. If we need to specialize the push we can do it in steps designed to implement that layer.
It is more straightforward to have in variants the source tag, the destination tag, and the namespace. even though you have to repeat some configuration values the code will be much simpler and easy to configure.

I understand that the configuration is the one we need to agree, and what do you propose?

vars/pushDockerImages.txt
kuisathaverat: I do not think so, you do not know which Docker image is the `source` and which will be the `destination` if you do not check the code.

vars/pushDockerImages.groovy
v1v: I don't think that's an issue:

`"${registry}/${sourceNamespace}/${name}${variant}:${sourceTag}"` then variable names should be defined with the format needed, otherwise what will you use?

v1v:
See https://github.com/elastic/apm-pipeline-library/pull/1550/commits/e187d6444cec9c0a94f3b36a80ba24a5f1526252

vars/pushDockerImages.txt
```diff
@@ -0,0 +1,67 @@
...
+    arch: 'amd64',
+    version: '8.2.0',
+    snapshot: true,
+    imageName: 'filebeat',
+    variants: [
+      '' : 'beats',
+      '-ubi8' : 'beats',
+      '-cloud' : 'beats-ci',
+      '-complete' : 'beats',
+    ],
```
mdelapenya, line 24: What is the contract to define variant keys and values? I know it's something the consumer should know what to put here, but for having an overall reference, as I'm not able to infer any pattern with the current example

v1v: The existing contract was already defined in the `packaging.groovy` as code:
- https://github.com/elastic/beats/blob/452d447372aac5449cf66e05c6e0eaaaf468dd66/.ci/packaging.groovy#L316-L326

This particular `variants` map provides:
- new variant name to be created and what targetNamespace.

Maybe the name could be changed to `variantsNamespaceMap` if it helps.

I still think this particular implementation could be much simpler if it's done by the build system itself. then from the CI point of view we only need to:

```
dockerLogin(...)
sh 'mage publishDockerImages'
```

For each beats.

vars/pushDockerImages.groovy
mdelapenya: I'd prefer hiding the `-` prefix, and let it as an implementation detail:

```groovy 
variants: [
      '' : 'beats',
      'ubi8' : 'beats',
      'cloud' : 'beats-ci',
      'complete' : 'beats',
    ],
```
and then evaluate it internally when building the image name: if the variant key is not empty, prepend the `-`. Otherwise the consumer could be adding whatever they like as variant separator, like `~cloud` or even worst `:cloud`. I know this is weird, but I'd see it better if we are controlling the inputs. We could even sanitise them, removing `:,\/;` etc...

Wdyt?

v1v: I understand the concern, but I'd rather prefer the freedom in case there is a corner case to support something else. No strong opinion though

vars/pushDockerImages.groovy
v1v: I understand that the configuration is the one we need to agree, and what do you propose?

vars/pushDockerImages.groovy
mdelapenya: Not a blocker on my side either. We have it already working in the old system and this PR is simply moving it to another shareable place 😄 I'm totally OK with keeping it as is.

vars/pushDockerImages.groovy
```diff
@@ -0,0 +1,132 @@
...
+      '-cloud' : 'beats-ci',
+      '-complete' : 'beats',
+    ],
+    targetNamespace: 'observability-ci'
+  )
+
+*/
+def call(Map args = [:]) {
+  if(!isUnix()){
+    error('publishToCDN: windows is not supported yet.')
```
v1v, line 46:
```suggestion
    error('pushDockerImages: windows is not supported yet.')
```

v1v:
> It is more straightforward to have in variants the source tag, the destination tag, and the namespace. even though you have to repeat some configuration values the code will be much simpler and easy to configure.

Do you mean something like the below snippet?

```
  pushDockerImages(
    secret: "my-secret",
    version: '8.2.0',
    snapshot: true,
    variantsMatrix: [
      [ source: "my-registry/beats/filebeat", arch: 'amd64', target: "my-registry/observability-ci/filebeat"]
      [ source: "my-registry/beats/filebeat-ubi8", arch: 'amd64', target: "my-registry/observability-ci/filebeat-ubi8"],
      [ source: "my-registry/beats-ci/filebeat-cloud", arch: 'amd64', target: "my-registry/observability-ci/filebeat-cloud"],
      [ source: "my-registry/beats-ci/filebeat-complete", arch: 'amd64', target: "my-registry/observability-ci/filebeat-complete"]
    ]
  )
```

mdelapenya (approved):

vars/README.md
```diff
@@ -194,9 +195,11 @@ Builds the Docker image for Kibana, from a branch or a pull Request.
 ```
 buildKibanaDockerImage(refspec: 'main')
 buildKibanaDockerImage(refspec: 'PR/12345')
+buildKibanaDockerImage(refspec: 'cf25ac3d1f8edff8f20003add707bfdc85d89fff', depth: 10)
+
 ```
 
-* refspec: A branch (i.e. main), or a pull request identified by the "pr/" prefix and the pull request ID.
+* refspec: A branch (i.e. main), a commit SHA, a tag, or a pull request identified by the "pr/" prefix and the pull request ID.
```
kuisathaverat, line 202: Tags are not supported, for release tags you can use the official image and for other tags you can use the commit SHA

```suggestion
* refspec: A branch (i.e. main), a commit SHA, or a pull request identified by the "pr/" prefix and the pull request ID.
```

kuisathaverat: I will fix it in other PR is not related with this one

v1v: This is unrelated to this PR, since it was generated, let's fix the comment in a follow up

PR #1558, opened 2022-02-17
feat: speed up checkouts

kuisathaverat:
## What does this PR do?

* speed up checkouts
* 
## Why is it important?

Changing the git fetch references to use only the referenced passed changes checkout of about 10 min to 1.5 min.

### DIFF
--- a/vars/fastCheckout.groovy
+++ b/vars/fastCheckout.groovy
@@ -1,0 +1,74 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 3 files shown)


vars/fastCheckout.groovy
```diff
@@ -0,0 +1,75 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+
+  Builds the Docker image for Kibana, from a branch or a pull Request.
```
v1v, line 20:
```suggestion
  Only make a Git checkout of the refspec passed as parameter, and not tags, this make the checkout faster.
This checkout does not work with tags.
```

v1v (approved):
Minor comment in the description

mdelapenya (approved):

PR #1560, opened 2022-02-22
Ensure null values are captured

v1v:
## What does this PR do?

Fix null values

## Why is it important?

Help with debugging since it fails but it's hard to know what's the genuine reason


vars/withGCPEnv.groovy
```diff
@@ -26,7 +26,7 @@ def call(Map args = [:], Closure body) {
   def credentialsId = args.containsKey('credentialsId') ? args.credentialsId : ''
   def secret = args.containsKey('secret') ? args.secret : ''
 
-  if (!credentialsId.trim() && !secret.trim()) {
```
v1v, line 29: Fixes 
image


then errors is only shown in the console output:

image

mdelapenya (approved):

PR #1575, opened 2022-03-02
bump automation for e2e-testing

v1v:
## What does this PR do?

Add e2e-testing

## Why is it important?

Then https://github.com/elastic/e2e-testing/pull/2198 won't be needed


.ci/.bump-stack-release-version.yml
```diff
@@ -49,3 +49,11 @@ projects:
     enabled: true
     labels: dependency
     title: "chore: update elastic stack release version"
+  - repo: e2e-testing
+    script: .ci/bump-stack-release-version.sh
+    branches:
+      - main
```
mdelapenya, line 55: Do we need main here? I expect to update the stale version for the last minor, only

v1v, line 55: This implementation does not use aliases, so we cannot use `8.<minor>` yet, so I'll hardcode this for the time being, then we can implement this feature, so I guess the last minor is `8.0` now?

```suggestion
      - "8.0"
```

v1v: @mdelapenya , what do you think?

mdelapenya: These are the branches we currently support: main, 8.1, 8.0 and 7.17.

OTOH, there is no `.ci/bump-stack-release-version.sh` file in the repo, is this something we need to add?

v1v: Yes, we need to do so, I guess the changes in https://github.com/elastic/e2e-testing/pull/2198 could be moved to the script `.ci/bump-stack-release-version.sh`

what do you think?

mdelapenya (approved):
We only need to check what versions need to run this automation. In the meantime LGTM

.ci/.bump-stack-release-version.yml
```diff
@@ -49,3 +49,13 @@ projects:
     enabled: true
     labels: dependency
     title: "chore: update elastic stack release version"
+  - repo: e2e-testing
+    script: .ci/bump-stack-release-version.sh
+    branches:
+      - "8.1"
```
v1v, line 55:
```suggestion
      - "main"
      - "8.1"
```

.ci/.bump-stack-release-version.yml
```diff
@@ -49,3 +49,14 @@ projects:
...
     labels: dependency
     title: "chore: update elastic stack release version"
+  - repo: e2e-testing
+    script: .ci/bump-stack-release-version.sh
+    branches:
+      - "main"
+      - "8.1"
+      - "8.0"
+      - "7.17"
+      - "7.17"
```
v1v, line 59:
```suggestion
      - "7.17"
```

v1v, line 59:
```suggestion
```

cachedout (approved):

v1v:
As we are moving away from jenkins specific, I'll close this PR now

Issue #1443, opened 2021-12-05
[k8s] some steps use the Agent labels


v1v:
### What

Some steps don't work with the k8s pod templates

```
Unhandled arch in NODE_LABELS: apm-shared-apm-pipeline-library-mbp-pr-1385-15-dzkgl-vbzr-4vncp apm-shared_apm-pipeline-library-mbp_pr-1385_15-dzkgl
```

### Why

An opinionated approach to know the arch, os and so on given an agent worker.

### Proposal

1) Add labels to the k8s agent
2) Support `uname` or another alternative to know the OS stuff.

v1v:
WE are moving to GH actions so no need work on this now

PR #1593, opened 2022-03-08
Support node labels for k8s pods

v1v:
## What does this PR do?

If `POD_LABEL` then set default behaviours

## Why is it important?

Support for k8s pods 

## Related issues
Closes https://github.com/elastic/apm-pipeline-library/issues/1443

### DIFF
--- a/vars/nodeOS.groovy
+++ b/vars/nodeOS.groovy
@@ -25,7 +25,7 @@ def call() {
   def labels = env.NODE_LABELS?.toLowerCase()
   def matches = []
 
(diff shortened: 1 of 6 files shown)


vars/nodeOS.groovy
```diff
@@ -37,6 +37,7 @@ def call() {
     matches.add('darwin')
   }
 
+
```
v1v, line 40:
```suggestion
```

kuisathaverat (approved):

PR #1594, opened 2022-03-08
jobdsl: add integrations@k8s eval

v1v:
## What does this PR do?

Add JobDSL for the https://github.com/elastic/integrations

## Why is it important?

Evaluate the k8s performance

### DIFF
--- a/.ci/jobDSL/jobs/fleet-ci/k8s/integrations_mbp.groovy
+++ b/.ci/jobDSL/jobs/fleet-ci/k8s/integrations_mbp.groovy
@@ -1,0 +1,116 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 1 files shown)


kuisathaverat (approved):

.ci/jobDSL/jobs/fleet-ci/k8s/integrations_mbp.groovy
```diff
@@ -0,0 +1,116 @@
...
+  }
+  branchSources {
+    branchSource {
+      source {
+        github {
+          id('k8s-evaluation/integrations-mbp') // IMPORTANT: use a constant and unique identifier
+          credentialsId('2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken')
+          repoOwner('v1v')
+          repository('integrations')
+          repositoryUrl('https://github.com/v1v/integrations.git')
```
cachedout, line 36: @v1v Is this pipeline still in development using your fork or should this have been swung over to the production repo?

cachedout (commented):
See inline comment about `repositoryUrl` on L36

.ci/jobDSL/jobs/fleet-ci/k8s/integrations_mbp.groovy
v1v: yes, we have planned to stress the k8s implementation in one of our productions instances but using our forked repos.
Those upstream repositories are already available in other CI controllers, and by using our forks we won't cause any harm in the other CI controllers, spam or anything similar, while we can enable the `cron` to run every x minutes, while using the upstream might produce duplicated builds in other CI controllers. 

This is only temporary

PR #1601, opened 2022-03-16
getBranchesFromAliases : support minor index alias

v1v:
## What does this PR do?

Support `major.<minor-$index>`
The index will be used to retrieve a previous minor version of the stack

## Why is it important?

It could be the case where there are multiple minors in development at the same time, 7.16 and 7.17, and with previous behaviour, the automation discarded those cases where the project needed a bump in two minors (See e2e-tests, that needs bumps in 7.16 and 7.17)

## Related issues

Similar to https://github.com/elastic/apm-pipeline-library/pull/1503 and notifies https://github.com/elastic/e2e-testing/pull/2241

### DIFF
--- a/vars/getBranchesFromAliases.groovy
+++ b/vars/getBranchesFromAliases.groovy
@@ -41,13 +41,14 @@ def call(Map args = [:]) {
 }
 
 def getBranchNameFromAlias(alias) {
-  // special macro to look for the latest minor version -1
-  if (alias.contains('8.<minor-1>')) {
-   return subtractBranchIfPossible(bumpUtils.getMajorMinor(bumpUtils.getCurrentMinorReleaseFor8()), 1)
+  // special macro to look for the latest minor version - subtrahend
+  if (alias.contains('8.<minor-')) {
+    def minorParts = alias.split('-')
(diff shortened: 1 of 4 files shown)


kuisathaverat (approved):

vars/getBranchesFromAliases.groovy
```diff
@@ -60,3 +65,28 @@ def getBranchNameFromAlias(alias) {
   }
   return alias
 }
+
+def subtractBranchIfPossible(String branch, String subtrahend) {
```
mdelapenya, line 69: If there is no collision with other method (i.e. one in which it's always possible), I'd prefer a simpler name
```suggestion
def subtractBranch(String branch, String subtrahend) {
```

v1v: No strong opinion. My aim was to be sure the name could clarify the expectation, it's not a normal subtraction, since the outcome could contain a negative minor, and if so, it will return the branch by default.

Let me rename it

mdelapenya: Not a blocker on my side though :D

mdelapenya (approved):

PR #1605, opened 2022-03-16
fix tar with 7z without extension

v1v:
## What does this PR do?

`7z` added the `7z` extension.

## Why is it important?

Fixes 

image

## Test

image

versus

image


mdelapenya (approved):

v1v:
I've just verified this is now working like a charm:

image

PR #1627, opened 2022-03-28
releaseManagerAnalyser and releaseManagerNotification step

v1v:
## What does this PR do?

Centralise the releaseManagerAnalyser and releaseManagerNotification steps

## Why is it important?

Help consumers to use the pipeline step then they don't need to worry about implementing anything but using the releaseManagerNotification step to notify and enrich the details with the analyser output

## Related issues

Closes https://github.com/elastic/apm-server/pull/7663 as it will use the new step instead

## Test

image

image

image

### DIFF
--- a/vars/releaseManagerAnalyser.txt
+++ b/vars/releaseManagerAnalyser.txt
@@ -1,17 +1,9 @@
-Given the release manager output then it analyseS the failure if any, and provides
-an environment variable, `DIGESTED_MESSAGE`, with the digested output to the end user.
+Given the release manager output then it analyses the failure if any, and returns
+the digested output to the end user.
 
 ```
-
 // analyse the release manager build output
-releaseManagerAnalyser(file: 'release-manager.out')
-
-// send an email\slack with the diggested message
-releaseNotification(slackChannel: "${env.SLACK_CHANNEL}",
-                      slackColor: args.slackStatus,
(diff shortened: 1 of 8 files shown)


vars/releaseManagerAnalyser.txt
```diff
@@ -0,0 +1,18 @@
+Given the release manager output then it analyseS the failure if any, and provides
+an environment variable, `DIGESTED_MESSAGE`, with the digested output to the end user.
+
+```
+
+// analyse the release manager build output
+releaseManagerAnalyser(file: 'release-manager.out')
+
+// send an email\slack with the diggested message
+releaseNotification(slackChannel: "${env.SLACK_CHANNEL}",
```
mdelapenya, line 10: Wdyt about combining both actions into one? Could it be error-prone having to remember how to pass the DIGESTED_MESSAGE to the next step?

Not a blocker, though

vars/releaseManagerAnalyser.txt
```diff
@@ -0,0 +1,18 @@
+Given the release manager output then it analyseS the failure if any, and provides
```
cachedout, line 1:
```suggestion
Given the release manager output then it analyses the failure if any, and provides
```

cachedout (approved):

mdelapenya (approved):

PR #1628, opened 2022-03-29
Add releaseManager step

v1v:
## What does this PR do?

Centralise the release manager step in a shared library

## Why is it important?

Simplify the consumers since they will just call a step in the pipeline. This step is not needed to run locally at all.

## Further details

I decided to use a script with env variables rather than using a script with parameters since it will run within the jenkins step and it reduces the number of steps to be able to use it, from

```
  def resourceContent = libraryResource("my-script")
  writeFile file: scriptFile, text: resourceContent
  sh(script: """#!/bin/bash +x
            chmod 755 my-script
            ./my-script param1 param2""")
```

to

```
  withEnv(["PARAM1=XXX", "PARAM2=YYY"]) {
    sh(script: libraryResource('my-script'))
  }
```

### Test

image

### DIFF
--- a/vars/releaseManager.txt
+++ b/vars/releaseManager.txt
@@ -1,0 +1,19 @@
+Given the project, release type and version it runs the release manager
+
+```
+// release a snapshot 8.2.0-SNAPSHOT for the APM Server and pick the files from build/dist
+releaseManager(project: 'apm-server',
+               version: '8.2.0-SNAPSHOT',
+               type: 'snapshot',
+               artifactsFolder: 'build/dist')
+
+// release a staging 8.2.0 for the APM Server
+releaseManager(project: 'apm-server',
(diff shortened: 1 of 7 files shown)


vars/releaseManager.txt
```diff
@@ -0,0 +1,20 @@
+Given the project, release type and version it runs the release manager
+
+```
+// release a snapshot 8.2.0-SNAPSHOT for the APM Server and pick the files from build/dist
+releaseManager(project: 'apm-server',
+               version: '8.2.0-SNAPSHOT',
```
v1v, line 6:
```suggestion
               version: '8.2.0',
```

kuisathaverat (approved):

mdelapenya (approved):

PR #1543, opened 2022-02-11
chore: use new junit2otel step

mdelapenya:
## What does this PR do?
It uses the `junit2otel` new step in this pipeline, setting the `JUNIT_2_OTLP` env var at the top-level environment section of the pipeline.

Because we are not passing service name/version or trace name, the step will infer them from:

- REPO as service name and trace name
- service version will be calculated from:
    - env.CHANGE_ID on PRs
    - env.TAG_NAME on tags
    - env.BRANCH_NAME on branches
    - `unknown` in any other case

We may need to improve this logic to automatically detect the project type:
- is it a Maven project? read the version from the pom.xml
- ...

## Why is it important?
Let's start collecting metrics from test results!

PR #1629, opened 2022-03-30
Update junit2otel step

mdelapenya:
- chore: do not protect the step with a env var
- chore: do not pass environment variables, use attributes

## What does this PR do?
It uses attributes instead of env vars when invoking the junit2otel step.

## Why is it important?
Consumers would need to use the step and its params, not setting up the environment accordingly, which is easier for them to mantain.

## Related issues
Completes #1543

### DIFF
--- a/vars/junit2otel.groovy
+++ b/vars/junit2otel.groovy
@@ -31,7 +31,6 @@ populated later on, using the https://github.com/mdelapenya/junit2otel library.
                     always {
                         // JUnit with OpenTelemetry traces
                         withEnv([
-                          "JUNIT_2_OTLP=true",
                           "OTEL_SERVICE_NAME=apm-pipeline-library",
                           "JUNIT_OTEL_SERVICE_VERSION=main",
                           "JUNIT_OTEL_TRACE_NAME=junit-tests"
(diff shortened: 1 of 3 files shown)


vars/junit2otel.groovy
```diff
@@ -55,33 +54,29 @@ populated later on, using the https://github.com/mdelapenya/junit2otel library.
...
+    error('junit2otel: docker is not installed but required.')
+  }
 
-    withOtelEnv() {
-      def testResults = args.containsKey('testResults') ? args.testResults : error("junit2otel: testResults parameter is required.")
-      def serviceName = env.JENKINS_OTEL_SERVICE_NAME?.trim() ? env.JENKINS_OTEL_SERVICE_NAME?.trim()+"-junit" : (env.REPO?.trim() ? env.REPO?.trim() : 'junit2otel')
-      def serviceVersion = env.JUNIT_OTEL_SERVICE_VERSION?.trim() ? env.JUNIT_OTEL_SERVICE_VERSION?.trim() : calculateFallbackServiceVersion()
-      def traceName = env.JUNIT_OTEL_TRACE_NAME?.trim() ? env.JUNIT_OTEL_TRACE_NAME?.trim() : (env.REPO?.trim() ? env.REPO?.trim() : 'junit2otel')
+  withOtelEnv() {
+    def testResults = args.containsKey('testResults') ? args.testResults : error("junit2otel: testResults parameter is required.")
```
v1v, line 66: By doing so, then https://github.com/elastic/apm-pipeline-library/pull/1629/files#diff-9de9abc4e2e6b84ce6d6cda7b44f73c8d7a66722272ef624c330ecbee01b9068R83 should be changed with the `args` but without `serviceName`...

Otherwise, the junit(args) will fail with the signature

mdelapenya: Done in 6e17ef63631d636892caf7bbcb08fc823559325d

PR #1638, opened 2022-04-01
Remove build override step

v1v:
## What does this PR do?

Remove the build override step, I was aimed to do it long time ago as it's only done to help with the UI click on buttons, since a known bug with BO view [JENKINS-56562](https://issues.jenkins-ci.org/browse/JENKINS-56562), but I'd rather prefer to keep our code simpler even if the UI is not nicer for the shake of simplicity and bug-free

## Why is it important?

Keep our code simpler, in addition, it removes the chances that the try/catch produces unexpected behaviours in the pipeline

image

## Related issues

Initially implemented in https://github.com/elastic/apm-pipeline-library/pull/259


kuisathaverat (approved):

mdelapenya (approved):

cachedout (approved):
💯

PR #1641, opened 2022-04-04
Fetch latest available release for the 8 major

v1v:
## What does this PR do?

Update current and next patch for the 8 Major only when the release is available

## Why is it important?

Avoid bumping the version when the release is not available yet, therefore https://github.com/elastic/apm-pipeline-library/pull/1639 won't be created since `8.2.0` has not been released.

This is a quite opinionated approach

### DIFF
--- a/.ci/updateStackReleaseVersion.groovy
+++ b/.ci/updateStackReleaseVersion.groovy
@@ -83,7 +83,7 @@ def fetchVersions() {
   // To store all the latest release versions
   def latestVersions = artifactsApi(action: 'latest-versions')
   def current7 = latestReleaseVersions.findAll { it ==~ /7\.\d+\.\d+/ }.sort().last()
-  def current8 = latestReleaseVersions.findAll { it ==~ /8\.\d+\.\d+/ }.sort().last()
+  def current8 = getCurrent8(latestReleaseVersions)
   // NOTE: 6 major branch is now EOL (we keep this for backward compatibility)
   releaseVersions[bumpUtils.current6Key()] = '6.8.23'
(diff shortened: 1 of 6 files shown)


mdelapenya (approved):

kuisathaverat:
the current 8.2.0 artifact is 8.2.0-BC1, when the Bc is available there are no new SNAPSHOTs, this will cause build using the SNAPSHOT no test the latest changes for a few weeks

v1v:
> the current 8.2.0 artifact is 8.2.0-BC1, when the Bc is available there are no new SNAPSHOTs, this will cause build using the SNAPSHOT no test the latest changes for a few weeks

I don't understand the issue. 

This change is only related to current releases and upcoming ones:

1) The `current8` should always point to an existing release.
2) The `nextMinor8` is the one to be pointing to `main`, but this change does nothing about this. 


To illustrate the above:
- https://github.com/elastic/apm-pipeline-library/pull/1639 was created without this change, and the semantic around `current_8` was not correct, since it should be `8.1.0`
- https://github.com/elastic/apm-pipeline-library/pull/1642/files bumped the version for `next-minor`

I only see an issue regarding `8.2.0` since it's not available in the current three aliases:
- current_8
- next_minor_8
- next_patch_8

image

Maybe `next_patch_8` should point to `8.2.0`  while `8.2.0` is still a `BC`? It might not be a quite accurate in terms of the semantic. Or we might need to revisit the existing semantic about current, next-(`minor`|`patch`)-`[0-9]` to support the BC

kuisathaverat (approved):

PR #1665, opened 2022-04-21
Support arguments for the coverageReport step

v1v:
## What does this PR do?

Add Map arguments for the coverageReport

## Why is it important?

Allow to customise the coverage file and location

## Related issues

Notifies https://github.com/elastic/observability-test-environments/pull/2242

### DIFF
--- a/src/test/groovy/CoverageReportStepTests.groovy
+++ b/src/test/groovy/CoverageReportStepTests.groovy
@@ -17,6 +17,7 @@
 
 import org.junit.Before
 import org.junit.Test
+import static org.junit.Assert.assertTrue
 
 class CoverageReportStepTests extends ApmBasePipelineTest {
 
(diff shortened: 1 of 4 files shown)


kuisathaverat (approved):

resources/scripts/coverage.sh
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env bash
```
mdelapenya, line 1: What is calling this script? I haven't seen a reference yet

mdelapenya (approved):

resources/scripts/coverage.sh
v1v: Ups! I pushed the wrong file to the wrong branch

v1v: I just reverted the last three commits that I pushed wrongly

PR #1685, opened 2022-05-11
GitHub build comment should not report syntax error if there are test errors

v1v:
## What does this PR do?

If there are test errors then it should not be reported as a syntax error

## Why is it important?

Avoid misleading

image

### DIFF
--- a/resources/github-comment-markdown.template
+++ b/resources/github-comment-markdown.template
@@ -115,9 +115,8 @@ ${description}
   </details>
 <%} else if (!buildStatus?.equals('ABORTED')) {%>
 ### Pipeline error [![1](https://img.shields.io/badge/1%20-red)](${env?.BUILD_URL}/console)
-  <p>
-  > This error is likely related to the pipeline itself. Click <a href="${env?.BUILD_URL}console">here</a>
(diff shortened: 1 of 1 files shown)


resources/github-comment-markdown.template
```diff
@@ -113,11 +113,10 @@ ${description}
   <%}%>
   </p>
   </details>
-<%} else if (!buildStatus?.equals('ABORTED')) {%>
+<%} else if (!buildStatus?.equals('ABORTED') && testsSummary?.failed == 0) {%>
```
mdelapenya, line 116: If there are test errors, shouldn't it be:
```suggestion
<%} else if (!buildStatus?.equals('ABORTED') && testsSummary?.failed > 0) {%>
```

v1v: Use Case:

- If there are test failures then do no report a syntax error regardless of the build status.
- If no test errors and build status different than `aborted` then report a syntax error

I guess `(!buildStatus?.equals('ABORTED') && testsSummary?.failed == 0) ` is the one supporting the above use cases, right?

mdelapenya: Ahh thanks for the clarification. Reading the title made me thought that way

PR #1698, opened 2022-05-18
withTerraformEnv step

v1v:
## What does this PR do?

Add withTerraformEnv step.

## Why is it important?

Refactor Beats and the apm-agent-dotnet

## Follow ups

If the tool is installed and matches with the given version the report about it. This should help to use the installation from the worker if the version matches.

## Test

https://github.com/elastic/apm-agent-dotnet/pull/1707
https://github.com/elastic/beats/pull/31663

### DIFF
--- a/vars/withTerraformEnv.groovy
+++ b/vars/withTerraformEnv.groovy
@@ -42,6 +42,9 @@ def downloadAndInstall(where, version) {
   dir(where) {
     download(url: url, output: zipfile)
     unzip(quiet: true, zipFile: zipfile)
+    if (isUnix()) {
+      sh(label: 'chmod terraform', script: 'chmod +x terraform')
+    }
   }
 }
 
--- a/vars/withTerraformEnv.groovy
+++ b/vars/withTerraformEnv.groovy
@@ -52,7 +52,7 @@ def terraformURL(version) {
   def url = 'https://releases.hashicorp.com/terraform'
   def arch = is64() ? 'amd64' : '386'
   if (isArm()) {
-    arch = 'arm64'
+    arch = is64() ? 'arm64' : 'arm'
   }
   return "${url}/${version}/terraform_${version}_${nodeOS()}_${arch}.zip"
 }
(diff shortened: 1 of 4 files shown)


vars/withTerraformEnv.groovy
```diff
@@ -0,0 +1,55 @@
...
+  dir(where) {
+    download(url: url, output: zipfile)
+    unzip(quiet: true, zipFile: zipfile)
+  }
+}
+
+def terraformURL(version) {
+  def url = 'https://releases.hashicorp.com/terraform'
+  def arch = is64() ? 'amd64' : '386'
+  if (isArm()) {
```
mdelapenya, line 51: This is because there is no arm 32 bits, right? Otherwise, we will be overriding that value

mdelapenya: Mmm terraform supports arm 32 bits here: 

[terraform_1.1.9_linux_arm.zip](https://releases.hashicorp.com/terraform/1.1.9/terraform_1.1.9_linux_arm.zip)
[terraform_1.1.9_linux_arm64.zip](https://releases.hashicorp.com/terraform/1.1.9/terraform_1.1.9_linux_arm64.zip)

pazone: Perhaps someone runs terraform on Raspberry Pi

cachedout (commented):
I find the naming and overall ergonomics of this to be a bit surprising. 

All this is really doing is just intstalling Terraform if it isn't already there but the name leads me to believe that it's some type of scoped environment for Terraform where that scoping will end when the block ends. 

Would it make more sense to just rename this to something like installTerraform() and not have accept a closure?

vars/withTerraformEnv.groovy
v1v: > This is because there is no arm 32 bits, right? Otherwise, we will be overriding that value

CI workers for ARM are 64 based, so I didn't support it since there is no consumers in our end for that, but let me add it so, whoever uses this library can benefit from

kuisathaverat (approved):

cachedout (approved):

PR #1705, opened 2022-05-23
runWatcher: subject with test status

v1v:
### What

Use status in the subject to highlight whether there were test failures in the last 7 days.

```gherkin
When there are not failing errors
Then the email should change to:
> Subject: $DATE: Last week Top failing Beats tests - 🚀 (no failures)
> Body: Over the past 7 days, these are the top 10 failing tests:
```

```gherkin
When there are failing errors
Then the email should change to:
> Subject: $DATE: Last week Top failing Beats tests - 💔 (test failures)
> Body: Over the past 7 days,  these are the top 10 failing tests:
```

### Why

Update the readability/usability of the email thread

### Implementation details

I decided to use - 💔  and 🚀 to highlight what happened rather than pointing to the number of matches with test failures. Since if even with 1 test failure is still a genuine information.

### Watcher

image

image

### DIFF
--- a/vars/runWatcher.groovy
+++ b/vars/runWatcher.groovy
@@ -79,9 +79,6 @@ def call(Map args = [:]) {
 }
 
 def getStatusMessage(parsed) {
-  def status = '🚀'
-  if (parsed?.watch_record.result?.input?.payload?.hits) {
-    status = parsed?.watch_record.result?.input?.payload?.hits.total > 0 ? '💔 (test failures)' : '🚀 (no failures)'
-  }
+  def status = parsed?.watch_record.result?.input?.payload?.hits?.total > 0 ? '💔 (test failures)' : '🚀 (no failures)'
   return status
 }


kuisathaverat (approved):

vars/runWatcher.groovy
```diff
@@ -76,3 +77,11 @@ def call(Map args = [:]) {
         log(level: 'WARN', text: "runWatcher: Didn't work")
     }
 }
+
+def getStatusMessage(parsed) {
+  def status = '🚀'
+  if (parsed?.watch_record.result?.input?.payload?.hits) {
```
mdelapenya, line 83: In which cases this is `false`? If so, the status will be 🚀 without any text.

mdelapenya (approved):
💯

vars/runWatcher.groovy
```diff
@@ -76,3 +77,11 @@ def call(Map args = [:]) {
...
+    status = parsed?.watch_record.result?.input?.payload?.hits.total > 0 ? '💔 (test failures)' : '🚀 (no failures)'
+  }
+  return status
```
v1v, line 86:
```suggestion
  def status = parsed?.watch_record.result?.input?.payload?.hits?.total > 0 ? '💔 (test failures)' : '🚀 (no failures)'
  return status
```

vars/runWatcher.groovy
v1v: I'll use the `?` so we can only support two cases, `total > 0` or otherwise no failed tests

PR #1711, opened 2022-05-26
golang: onboard elastic-agent-inputs

v1v:
## What does this PR do?

onboard elastic-agent-inputs for the golang autobump, it uses the current golang version `1.18` rather than `1.17`

## Related issues

Uses https://github.com/elastic/elastic-agent-inputs/pull/13

Closes https://github.com/elastic/apm-pipeline-library/issues/1712

### DIFF
--- a/.ci/.bump-go-release-version.yml
+++ b/.ci/.bump-go-release-version.yml
@@ -56,6 +56,12 @@ projects:
     overrideGoVersion: "1.17"
     enabled: true
     labels: dependency
+  - repo: elastic-agent-inputs
+    script: .ci/bump-go-release-version.sh
+    branches:
+      - main
+    enabled: true
+    labels: dependency
   - repo: elastic-agent-libs
     script: .ci/bump-go-release-version.sh
     branches:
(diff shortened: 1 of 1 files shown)


.ci/.bump-go-release-version.yml
```diff
@@ -56,6 +56,12 @@ projects:
     overrideGoVersion: "1.17"
     enabled: true
     labels: dependency
+  - repo: elastic-agent-inputs
+    script: .ci/bump-go-release-version.sh
+    branches:
+      - main
+    enabled: true
```
v1v, line 63:
```suggestion
    overrideGoVersion: "1.17"
    enabled: true
```

v1v: Only if `1.17` is the minor Golang version, otherwise it will point to the latest one, so `1.18` now

kuisathaverat (approved):

.ci/.bump-go-release-version.yml
kvch: I would like to stay on 1.17 like in Beats.

kuisathaverat (approved):

PR #1728, opened 2022-06-14
Add withElasticsearchDeploymentEnv step

v1v:
## What does this PR do?

Refactor `withClusterEnv` that uses the new step called `withElasticsearchDeploymentEnv`, then I'll enable some other steps to configure the fleet and kibana env variables accordingly.

To help with we have agreed to use `Deployment` for the specific Elastic Stack, and `Cluster` for only the `withClusterEnv` for backward compatibility reasons, for instance: `withElasticsearchDeploymentEnv` rather than `withElasticsearchClusterEnv`

## Why is it important?

Refactor `withClusterEnv` to support more options

## Related issues
Requires https://github.com/elastic/apm-pipeline-library/pull/1727
Notifies https://github.com/elastic/apm-pipeline-library/pull/1729 and https://github.com/elastic/apm-pipeline-library/pull/1730

### DIFF
--- a/vars/withElasticsearchClusterEnv.groovy
+++ b/vars/withElasticsearchClusterEnv.groovy
@@ -1,0 +1,55 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+  Wrap the Elasticsearch cluster credentials and entrypoints as environment variables that are masked
+
+  withElasticsearchClusterEnv(cluster: 'test-cluster-azure') {
+    // block
+  }
+*/
+
+def call(Map args = [:], Closure body) {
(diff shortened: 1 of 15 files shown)


vars/withElasticsearchClusterEnv.txt
```diff
@@ -0,0 +1,11 @@
+  Wrap the Elasticsearch cluster credentials and entrypoints as environment variables that are masked
```
mdelapenya, line 1: Not a blocker, but I'd prefer using the word `deployment` instead of `cluster` to follow the same terminology than Elastic Cloud

vars/withElasticsearchClusterEnv.groovy
```diff
@@ -0,0 +1,55 @@
...
+  def es_url = es.url
+  def username = es.username
+  def password = es.password
+  validateField(es_url, "withElasticsearchClusterEnv: was not possible to get the authentication info for the url field.")
+  validateField(username, "withElasticsearchClusterEnv: was not possible to get the authentication info for the username field.")
+  validateField(password, "withElasticsearchClusterEnv: was not possible to get the authentication info for the password field.")
+  withEnvMask(vars: [
+    [var: 'ES_URL', password: es_url],
+    [var: 'ES_USERNAME', password: username],
+    [var: 'ES_PASSWORD', password: password]
```
mdelapenya, line 45: Wdyt about expanding the variables to also populate the `ELASTICSEARCH_` prefix? It would result in adding 6 variables instead of 3.

Therefore, products consuming those Elastic standard variables would directly benefit from it without doing a translation

vars/withElasticsearchClusterEnv.txt
v1v: for backward compatibility I'll keep `withClusterEnv` but change `withElasticsearchClusterEnv` for `withElasticsearchDeploymentEnv`, similar for `Kibana` and `Fleet`, is that ok?

mdelapenya: Perfect!

cachedout (approved):

PR #1729, opened 2022-06-14
Support kibana in the withClusterEnv

v1v:
## What does this PR do?

Support Kibana env variables when using the withClusterEnv (optional)

## Why is it important?

Backward compatible, so, consumers should not be affected, so the kibana is optional and unset by default.

## Related issues
Requires https://github.com/elastic/apm-pipeline-library/pull/1728 https://github.com/elastic/apm-pipeline-library/pull/1727


vars/withKibanaClusterEnv.groovy
```diff
@@ -0,0 +1,55 @@
...
+  def props = getVaultSecret(secret: secret)
+  if (props?.errors) {
+    error "withKibanaClusterEnv: Unable to get credentials from the vault: ${props.errors.toString()}"
+  }
+  def dataJson = props?.data.value
+  def kibana = toJSON(dataJson)
+  def kibana_url = kibana.url
+  def username = kibana.username
+  def password = kibana.password
+  validateField(kibana_url, "withKibanaClusterEnv: was not possible to get the authentication info for the url field.")
```
cachedout, line 39: Should the validation go inside the env masking below or will the validator mask them on its own?

v1v: `validateField` does not print anything, so it's a safe function as far as I see, or do you think there is another scenario that I missed?

cachedout: Should be OK then.

cachedout (approved):

PR #1737, opened 2022-06-16
Add isEmpty and errorIfEmpty

v1v:
## What does this PR do?

Add `isEmpty` and `errorIfEmpty`

## Why is it important?

Avoid duplicated code

### DIFF
--- a/vars/validateValue.groovy
+++ b/vars/validateValue.groovy
@@ -1,0 +1,26 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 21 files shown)


vars/validateValue.groovy
```diff
@@ -0,0 +1,26 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/**
+Given the value it validates if it's not null or empty otherwise it reports an error
```
cachedout, line 19: This is just a nit, but I don't think that this is a good name for this function because the word "validation" is very broad and can mean many things. To me, it would be better to name it something like `isValue()` or perhaps `notEmpty()`. WDYT?

kuisathaverat: I vote for `notEmpty` or `isBlank`

v1v: `notEmpty` then

v1v: I somehow decided to slightly diverge from `notEmpty` and use `isEmpty` and `errorIfEmpty` to avoid the negation and support a validation too, let me know what you think :)

kuisathaverat (approved):

vars/validateValue.groovy
cachedout: > let me know what you think

👍 that is better

PR #1739, opened 2022-06-17
Add ELASTICSEARCH env variables for the withElasticsearchDeploymentEnv

v1v:
## What does this PR do?

Add `ELASTICSEARCH_` env variables and keep support for the existing ones for backward compatibility reasons.

## Why is it important?

Those are the env variables used by some other projects, such as the e2e-testing and it seems they are the official ones

There is also support for `ELASTICSEARCH_HOSTS` but for the time being I'll not include

## Related issues

Notifies https://github.com/elastic/e2e-testing/pull/2626

### DIFF
--- a/vars/README.md
+++ b/vars/README.md
@@ -817,6 +817,13 @@ Get the stage ID in the current context.
 def stage = getStageId()
 ```
 
+## getTestClusterSecret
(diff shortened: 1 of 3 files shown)


mdelapenya (approved):
LGTM 👍 

image

PR #1743, opened 2022-06-22
Retry only for the vault access in withVaultToken

v1v:
## What does this PR do?

Retry logic only for the vault access in withVaultToken

## Why is it important?

Otherwise it runs the given body, but the retry was only aimed to add stability while accessing Vault.


mdelapenya (approved):
💯

kuisathaverat (approved):

Issue #1744, opened 2022-06-23
GitHub build comment failed


v1v:
<img width="1433" alt="image" src="https://user-images.githubusercontent.com/2871786/175258810-878da781-d1d9-4034-932a-e673b797898a.png">

https://beats-ci.elastic.co/blue/organizations/jenkins/e2e-tests%2Fe2e-testing-mbp/detail/PR-2688/1/pipeline/19094

```
[2022-06-23T07:04:47.882Z] ERROR: generateBuildReport: Error generating build report
[2022-06-23T07:04:47.883Z] groovy.lang.MissingMethodException: No signature of method: net.sf.json.JSONNull.contains() is applicable for argument types: (java.lang.String) values: [githubPrCheckApproved]
[2022-06-23T07:04:47.883Z] Possible solutions: toString(), toString(), toString(), notify(), toString(int)
[2022-06-23T07:04:47.883Z] 	at com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.findAll(CpsDefaultGroovyMethods.java:1068)
[2022-06-23T07:04:47.883Z] Caused: groovy.text.TemplateExecutionException: Template execution error at line 82:
[2022-06-23T07:04:47.883Z]         81:                                       !it?.displayName?.contains('Archive JUnit') &&
[2022-06-23T07:04:47.883Z]     --> 82:                                       !it?.displayDescription?.contains('githubPrCheckApproved') &&
[2022-06-23T07:04:47.883Z]         83:                                       !it?.displayDescription?.contains('approval-list/elastic') &&
[2022-06-23T07:04:47.883Z] 
[2022-06-23T07:04:47.883Z] 	at com.cloudbees.groovy.cps.sandbox.DefaultInvoker.methodCall(DefaultInvoker.java:20)
[2022-06-23T07:04:47.883Z] 	at com.cloudbees.groovy.cps.impl.ContinuationGroup.methodCall(ContinuationGroup.java:86)
[2022-06-23T07:04:47.883Z] 	at com.cloudbees.groovy.cps.impl.FunctionCallBlock$ContinuationImpl.dispatchOrArg(FunctionCallBlock.java:113)
[2022-06-23T07:04:47.883Z] 	at com.cloudbees.groovy.cps.impl.FunctionCallBlock$ContinuationImpl.fixName(FunctionCallBlock.java:78)
[...]

PR #1745, opened 2022-06-23
Fix JSONNull in the build notifications

v1v:
## What does this PR do?

fix generateBuildReport: net.sf.json.JSONNull

## Why is it important?

another corner case

UTs uses a different Json transformation versus the one used in the pipeline itself. 

https://github.com/jenkinsci/pipeline-utility-steps-plugin/blob/7c062f79bd041ede1692913a5df9182b7bf8b7af/src/main/java/org/jenkinsci/plugins/pipeline/utility/steps/json/ReadJSONStepExecution.java#L87

So I could not reproduce it locally until I changed to use the right Json transformation

## Related issues
Similar to https://github.com/elastic/apm-pipeline-library/pull/1617
## Related issues
Closes https://github.com/elastic/apm-pipeline-library/issues/1744


src/test/groovy/ApmBasePipelineTest.groovy
```diff
@@ -746,6 +747,15 @@ class ApmBasePipelineTest extends DeclarativePipelineTest {
     return jsonSlurper.parseText(jsonText)
   }
 
+  def readNetJSON(params){
```
mdelapenya, line 750: readNetJson? why `Net` in the name 😄

mdelapenya (approved):

src/test/groovy/ApmBasePipelineTest.groovy
v1v: It uses `net.sf.json.JSONSerializer` so I wanted a new test function so the tests could benefit from, but I didn't want to change the existing implementation since it might produce some errors in the existing UTs.

I can change it for something else, open for suggestions

mdelapenya: I'd name it `toJSON` and delegate the implementation details to the internals of the method

v1v: I'm afraid `toJSON` clashes with some existing calls with a different signature. so I tried to refactor it, but I'm afraid it's not that simple. I just renamed the function with something else most descriptive about what args are allowed.

src/test/groovy/NotificationManagerStepTests.groovy
```diff
@@ -1109,4 +1109,21 @@ class NotificationManagerStepTests extends ApmBasePipelineTest {
...
+      buildStatus: "FAILURE",
+      changeSet: [],
+      statsUrl: "https://ecs.example.com/app/kibana",
+      stepsErrors: textOrFiletoJSON(file: "1744/steps-errors.json"),
+      testsErrors: [:],
+      testsSummary: []
+    )
+    printCallStack()
+    assertJobStatusSuccess()
+    assertFalse(assertMethodCallContainsPattern('githubPrComment', 'message=}'))
```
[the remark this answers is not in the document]
v1v: Yes, I wanted to verify the message was empty, so the existing Unit Testing for Jenkins pipelines is a bit rudimentary in terms of the assertions, since it uses plain strings

src/test/groovy/NotificationManagerStepTests.groovy
```diff
@@ -1109,4 +1109,21 @@ class NotificationManagerStepTests extends ApmBasePipelineTest {
...
+    assertTrue(assertMethodCallContainsPattern('githubPrComment', 'Show only the first 10 steps failures'))
```
v1v, line 1127:
```suggestion
    assertTrue(assertMethodCallContainsPattern('githubPrComment', 'Show only the first 10 steps failures'))
```

cachedout (approved):

PR #1748, opened 2022-06-24
Generate report with the Go benchmark

v1v:
## What does this PR do?

* Add Go benchmark report for the `notifyBuildResult`. Only for reporting purposes
* Step to generate the Go benchmark report so it can be consumed by the `notifyBuildResult`.

## Why is it important?

Allow to run the Go benchmark on a PR basis and compare the output with the target branch.

## Implementation details

This is a bit opinionated as it required to generate some benchmark, run the diff and notify the result. So the contract relies on:
- the file with the benchmark, so it can be copied from the TARGET_BRANCH, then stash it to be later on consumed by the `notifyBuildResult`.

Then the `notifyBuildResult` step will unstash the file with the diff to be able to create the GitHub comment.

The existing framework to notify build results in the CI is done by the `notifyBuildResult`, so as consumer there is nothing else to do but use the new step called `generateGoBenchmarkDiff ` to generate the bench diff and enable the new argument in the `notifyBuildResult`.

I decoupled `generateGoBenchmarkDiff` from `notifyBuildResult`, since the benchmark report could happen in a different CI worker so the report file could be somewhere else.

## Test

`test/report-bench-diff` branch is equivalent to this PR, so I could run some tests in https://github.com/elastic/apm-server/pull/8471

https://github.com/elastic/apm-server/pull/8471#issuecomment-1165633680 was automatically reported:

image


vars/generateGoBenchmarkDiff.groovy
```diff
@@ -0,0 +1,84 @@
...
+}
+
+def runBenchmarkDiff(Map args = [:]) {
+  def diffReport = getReportFileName()
+  def flags = ''
+  if (args.filter.equals('exclude')) {
+    flags = "| grep -v 'all equal' | grep -v '~'"
+  }
+  withGoEnv(pkgs: [ "golang.org/x/perf/cmd/..." ] ) {
+    sh(label: 'generateGoBenchmarkDiff', script: "benchstat ${args.file} ${args.compareWith} ${flags} | tee ${diffReport}")
```
marclop, line 74: can we change the order of the arguments? in general `benchstat` is run with `benchstat old.result new.result` so that you get the right data in the right columns, otherwise we'll get the results backwards.

v1v: @marclop , I'm a bit confused with the `grep` filters:

```
benchstat 7.17.out 8.3.out | grep -v 'all equal' | grep -v '~' # Exclude unchanged results. 
name                                                old time/op    new time/op    delta
pkg:github.com/elastic/apm-server/agentcfg goos:linux goarch:amd64
pkg:github.com/elastic/apm-server/model/modelindexer goos:linux goarch:amd64
pkg:github.com/elastic/apm-server/processor/stream goos:linux goarch:amd64
pkg:github.com/elastic/apm-server/publish goos:linux goarch:amd64
pkg:github.com/elastic/apm-server/x-pack/apm-server/aggregation/spanmetrics goos:linux goarch:amd64
pkg:github.com/elastic/apm-server/x-pack/apm-server/aggregation/txmetrics goos:linux goarch:amd64
pkg:github.com/elastic/apm-server/x-pack/apm-server/sampling goos:linux goarch:amd64
pkg:github.com/elastic/apm-server/x-pack/apm-server/sampling/eventstorage goos:linux goarch:amd64
```

for such I downloaded the bench files for 7.17 and 8.3:

```
wget https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/7.17/301/artifact/src/github.com/elastic/apm-server/bench.out -O 7.17.out
wget https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/8.3/32/artifact/src/github.com/elastic/apm-server/bench.out -O 8.3.out  

```

Is the above the expected report output?

mdelapenya (approved):

vars/generateGoBenchmarkDiff.groovy
v1v: Do you mean `benchstat` should use the `bench.out` that was generated using those new env variables?

marclop: yes, we should modify the way we run the benchmarks, so they're run a few times (5) is a good place to start and then `benchstat` should use the newly generated `bench.out`.

marclop: I think we still need to modify `.ci/scripts/bench.sh` line 4 so it includes`BENCH_COUNT=5`. I didn't make that change in the PR I merged today, so that's still pending.

v1v: https://github.com/elastic/apm-server/pull/8476 is the one with the proposal and [build 2](https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-8476/2/) is the one that will run the benchmark for the PR, let's see how it goes

kuisathaverat (approved):

vars/notifyBuildResult.txt
```diff
@@ -40,6 +40,7 @@ emails on Failed builds that are not pull request.
 * aggregateComments: Whether to create only one single GitHub PR Comment with all the details. Default true.
 * jobName: The name of the job, e.g. `Beats/beats/main`.
 * notifyCoverageComment: Whether to add a comment in the PR with the coverage summary as a comment. Default: `true`.
+* notifyGoBenchmarkComment: Whether to add a comment in the PR with the go benchmark summary as a comment. Default: `false`. It requires the file name `goBenchmark.diff`
```
v1v, line 43:
```suggestion
* notifyGoBenchmarkComment: Whether to add a comment in the PR with the go benchmark summary as a comment. Default: `false`. It requires the file name `bench.diff` which it's generated by the `generateGoBenchmarkDiff` step.
```

PR #1757, opened 2022-06-30
automation: enable extended description for the Golang autobump

v1v:
## What does this PR do?

Enable to customise the PR description for the `golang` autobump.

## Why is it important?

@cmacknz asked about a particular PR that was failing -> https://github.com/elastic/beats/pull/31827#issuecomment-1168846705

This proposal could help with adding further context regarding the `golang-crossbuild` dependencies with Beats and Elastic Agent.

Currently

image

Proposal

image

The existing automation is stateless so, it does not look after cross-project dependencies, so by adding the description with more semantic, I'd expect to answer those questions, what do you think?

### DIFF
--- a/.ci/.bump-go-release-version.yml
+++ b/.ci/.bump-go-release-version.yml
@@ -12,6 +12,7 @@
 ##  assign: A comma-separated list (no spaces around the comma) of GitHub handles to assign to this pull request. Optional.
 ##  reviewer: A comma-separated list (no spaces around the comma) of GitHub handles to request a review from. Optional.
 ##  overrideGoVersion: The major.minor version to be used since some projects do not use the latest major.minor version but the previous one. Optional.
+##  description: Extended description to be added to the existing one in the PR. Optional.
 ##
 projects:
   - repo: apm-pipeline-library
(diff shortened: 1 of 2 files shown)


cmacknz (approved):
I like it! It makes the build failing for version bumps less mysterious.

.ci/.bump-go-release-version.yml
```diff
@@ -48,6 +50,7 @@ projects:
     overrideGoVersion: "1.17"
     enabled: true
     labels: dependency,backport-skip
+    description: It requires the version to be bumped first in the https://github.com/elastic/golang-crossbuild. Otherwise it will fail until the docker images are available.
```
v1v, line 53:
```suggestion
    description: It requires the version to be bumped first in golang-crossbuild project, then a new release will be added to https://github.com/elastic/golang-crossbuild/releases. Otherwise it will fail until the docker images are available.
```

.ci/.bump-go-release-version.yml
```diff
@@ -35,6 +36,7 @@ projects:
...
```
v1v, line 39:
```suggestion
    description: It requires the version to be bumped first in golang-crossbuild project, then a new release will be added to https://github.com/elastic/golang-crossbuild/releases. Otherwise it will fail until the docker images are available.
```

PR #1760, opened 2022-07-04
ci: skip default checkout

v1v:
## What does this PR do?

disable skip checkout since we use the `checkout/stash` pattern

## Why is it important?

Remove unrequired checkouts per stage

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -69,6 +69,7 @@ pipeline {
       }
     }
     stage('Check Licenses') {
+      options { skipDefaultCheckout() }
       steps {
         withGithubNotify(context: 'Check Licenses') {
           deleteDir()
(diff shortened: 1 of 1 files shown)


amannocci:
Since we need to skip the checkout step in every stage, maybe we could use a global `skipDefaultCheckout()` in `options` and remove all the others?

v1v:
> Since we need to skip the checkout step in every stage, maybe we could use a global skipDefaultCheckout() in options and remove all the others?

IIRC, we cannot use it since some environment variables won't be populated. A kind of tribal knowledge that I cannot track where we discussed it :/

I tried to reproduce the environmental issues in https://github.com/elastic/apm-pipeline-library/pull/1761 but I'm afraid I cannot, so maybe they are not anymore an issue!


@kuisathaverat do you recall what's the upstream issue with this particular behaviour?

kuisathaverat (approved):

kuisathaverat:
It is not possible to put the skipDefaultCheckout(), if you do that, some SCM environment variables won't be defined

amannocci (approved):

PR #1777, opened 2022-07-11
action: opentelemetry exporter

v1v:
## What does this PR do?

Enable the OpenTelemetry exporter for GitHub actions with JUnit support

## Why is it important?

Keep iterating on this so we can gather further insights and see how the existing integration with https://github.com/jenkinsci/opentelemetry-plugin could be also reused (attributes, dashboards) if possible.

## Important

For some reason, I cannot test the GitHub actions on this PR but only when it's merged, I don't know whether this will work, since we run some docker containers as part of the UTs.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,6 +1,6 @@
 name: Java CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,0 +1,26 @@
+name: Java CI
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          java-version: '11'
+          distribution: 'adopt'
+      - name: Build with Maven
+        run: mvn --batch-mode --update-snapshots package
+      - uses: v1v/otel-upload-test-artifact-action@v2
+        if: always()
+        continue-on-error: true
+        with:
+          jobName: "build"
+          stepName: "Build with Maven"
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,26 @@
...
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          java-version: '11'
+          distribution: 'adopt'
```
kuisathaverat, line 15: adopt JDK stop releasing new Docker images about 9 months ago https://hub.docker.com/_/eclipse-temurin seems the better replacement

v1v: I used the one defined in https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-maven#using-the-maven-starter-workflow

We might need to contribute also in the github docs :)

I'll update it with the proposal, although I cannot find the way to test the GitHub actions for the very first time until they are not merged into the target branch

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Build with Maven
+        run: mvn --batch-mode --update-snapshots package
+      - uses: v1v/otel-upload-test-artifact-action@v2
```
kuisathaverat, line 18: this action collects spans after the execution of the steps, it is fine, but maybe there is an implementation that uses [check_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#check_run) event trigger this would be more "real" span

kuisathaverat (approved):

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,26 @@
+name: Java CI
+
+on: [push]
```
v1v, line 3:
```suggestion
on: [push, pull_request]
```

PR #1775, opened 2022-07-08
ci: enable k8s workers

v1v:
## What does this PR do?

Initial attempt to use the `k8s` gobld workers in those pipelines which are not critical or likely platform specific.

## Why is it important?

Validate if we can start using it

## Test

I triggered manually some pipelines by forcing the `label` to `k8s`:

* https://apm-ci.elastic.co/job/apm-shared/job/apm-schedule-daily/812/console
* https://apm-ci.elastic.co/job/apm-shared/job/apm-schedule-weekly/187/console
```
15:21:45  [Pipeline] Start of Pipeline
15:21:45  [Pipeline] node
15:22:01  Running on [apm-ci-immutable-jenkins-swarm-1657290110773068510](https://apm-ci.elastic.co/computer/apm-ci-immutable-jenkins-swarm-1657290110773068510/) in /home/jenkins/workspace/apm-shared/apm-schedule-weekly
15:22:01  [Pipeline] {
```

* https://apm-ci.elastic.co/job/apm-shared/job/bump-go-release-version-pipeline/84/console

PR #1799, opened 2022-07-20
Prepare jenkins-agent with the required tools

v1v:
## What does this PR do?

Host the jenkins-agent docker image

## Why is it important?

Then we can add the required toolchain and use it when running Jenkins `k8s` workers with `Gobld`. For instance https://github.com/elastic/apm-pipeline-library/pull/1800 is a way to use this particular image

## Important

As long as the existing docker image uses centos we are forced to use the hack with the mirrors, already reported to the CI Systems team.

the jenkins-agent container requires some env variables, so the TAP UTs will fail, so `simple-test` contains a subset of tests to be executed

## Related issues

Caused by https://github.com/elastic/apm-pipeline-library/commits/main

### DIFF
--- a/.ci/docker/Makefile
+++ b/.ci/docker/Makefile
@@ -42,7 +42,11 @@ convert-tests-results: ## convert TAP test results to JUnit
 					sh -c 'npm install tap-xunit -g && cat target/results.tap | tap-xunit --package="co.elastic.pipeline.$${APP}" > target/junit-$${APP}-results.xml'
 
 test-%: prepare-test ## Run the tests for the specific app
-	@DOCKERFILE=$* bats-core/bin/bats --tap tests | tee target/results.tap
+	@DOCKERFILE=$* bats-core/bin/bats --tap tests/tests.bats | tee target/results.tap
+	@$(MAKE) -s convert-tests-results
+
+simple-test-%: prepare-test ## Run the specific tests for the specific app
+	@DOCKERFILE=$* bats-core/bin/bats --tap tests/tests_without_run.bats | tee target/results.tap
 	@$(MAKE) -s convert-tests-results
 
 push-%: prepare-test ## Push the Docker image to the docker.elastic.co repository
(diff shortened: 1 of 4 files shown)


kuisathaverat (approved):

cachedout:
I don't fully understand the reasoning behind this. Could you provide additional explanation?

v1v:
> I don't fully understand the reasoning behind this. Could you provide additional explanation?

The existing default k8s agents do not have all the required toolchain, the CI systems team provide a guideline about how to customise those docker images, in their example the used `ruby`.

In our case, if we use the default docker image provided by the `gobld` with k8s support, it will not have some of the tools we need. Therefore we won't be able to benefit from the k8s workers for some of the use cases.

For instance, https://apm-ci.elastic.co/job/apm-shared/job/update-stack-release-version-pipeline/204/ failed recently since `jq` was not installed. That particular pipeline was migrated to `k8s` with https://github.com/elastic/apm-pipeline-library/pull/1775

https://github.com/elastic/apm-pipeline-library/pull/1775#issuecomment-1189139356 was a comment regarding some of the pipelines did have a dependency with `jq`, I actually removed them from the original PR https://github.com/elastic/apm-pipeline-library/pull/1775/commits/27e87494dc7f8150d6c12bddbe35b6000a1922ee but somehow I missed the one for `update-stack-release-version-pipeline`

In any case, this particular PR is a follow up to use k8s specific workers, but using docker images with the toolchain required to run some of our pipelines.

Let me know if you need further details

cachedout:
@v1v Thanks for the explanation. Makes sense. Once this image is built and push, how do Jenkins instances consume it? Is it via a label?  Or is there just a single Docker image which is used for all jobs with the `ks8` label?

v1v:
> Once this image is built and push, how do Jenkins instances consume it? Is it via a label? Or is there just a single Docker image which is used for all jobs with the ks8 label?
With a label.

There are two approaches:
- Using some of the special labels provided by `gobld`, see https://github.com/elastic/apm-pipeline-library/pull/1775
- Using some of the static labels provided by `gobld`, that's stored in the `beats-ci.yml` or similar files, I proposed to use `k8s`  by default to consume this particular image, so there is no need to do much

cachedout (approved):

PR #1801, opened 2022-07-21
Trigger docker images more frequently

mdelapenya:
- chore: build and push stack images every hour
- fix: move pipeline to the right directory

## What does this PR do?
It runs the build every hour, also moving the pipeline file to the right folder in the workspace.

## Why is it important?
We want to have the published snapshots from Infra as soon as possible, so that we are able to consume them from other services, such as the oblt-cli, the obl-robot, etc


v1v (approved):

.ci/jobDSL/jobs/apm-ci/apm-shared/oblt-test-env/stack_docker_images.groovy
```diff
@@ -30,7 +30,7 @@ pipelineJob("apm-shared/oblt-test-env/stack-docker-images") {
     artifactDaysToKeep(-1)
   }
   triggers {
-    cron('H H(3-4) * * 1-5')
+    cron('0 */1 * * 1-5')
```
v1v, line 33: It's a cached approach? Otherwise, I'd say to run less often? Or run between a time range? I guess AMAC and EMEA are the interested ones, while APAJ might be a bit less used? Just an idea

kuisathaverat, line 33:
```suggestion
    cron('H H * * 1-5')
```

kuisathaverat:
```suggestion
    cron('0 * * * 1-5')
```

kuisathaverat:
```suggestion
    cron('@hourly')
```

mdelapenya: I understand that we want to do this even on weekends, right?

kuisathaverat (approved):

Issue #1808, opened 2022-07-25
action: get-changed-files gets an error 


v1v:
<img width="1060" alt="image" src="https://user-images.githubusercontent.com/2871786/180746886-d212b56a-2464-4b93-b69c-890667c70991.png">

https://github.com/elastic/apm-pipeline-library/pull/1807

v1v:
See https://github.com/jitterbit/get-changed-files/issues/10

PR #1809, opened 2022-07-25
actions: push on main and pull requests

v1v:
## What does this PR do?

Filter when to run actions (on PRs by default , and on branches on main only)

## Why is it important?

If a branch in the upstream with a PR, then it will run twice

## Related issues

Closes https://github.com/elastic/apm-pipeline-library/issues/1808


amannocci (approved):
LGTM

PR #1815, opened 2022-07-25
Use latest pre-commit & jitterbit/get-changed-files actions

amannocci:

## What does this PR do?

* Upgrade to the latest pre-commit action [v3.0.0](https://github.com/marketplace/actions/pre-commit).
* Upgrade to the latest jitterbit/get-changed-files action [v1](https://github.com/jitterbit/get-changed-files)

## Why is it important?

* Get the latest security and bug fix updates.


v1v (changes requested):
@kuisathaverat , asked to use fixed versions instead

amannocci:
Then, we are already good.

PR #1821, opened 2022-07-26
fix empty list of 7 releases

v1v:
## What does this PR do?

If the list of versions are empty then default to the previous value

## Why is it important?

Artifacts API might not provide what are the current releases as they are cleaned up after certain time...

## Test

I ran this [build](https://apm-ci.elastic.co/job/apm-shared/job/update-stack-release-version-pipeline/217/console) and it worked fine


cachedout (commented):
Tests?

v1v:
> Tests?

I tested the pipeline itself and it ran successfully. Or what do you mean?

cachedout:
> I tested the pipeline itself and it ran successfully. Or what do you mean?

Oops, ignore me. I was going to fast and thought this was part of the library itself and not the pipeline.

cachedout (approved):

PR #1822, opened 2022-07-27
k8s: add gvm for the e2e-testing

v1v:
## What does this PR do?

Add `gvm` to the docker image

## Why is it important?

Run on k8s the e2e-testing -> https://github.com/elastic/e2e-testing/pull/2859


v1v:
This is not needed since the e2e-testing requires docker for running the UTs.

PR #1833, opened 2022-08-08
Move from observablt-robots-on-call to observablt-robots

amannocci:

## What does this PR do?

Use `elastic/observablt-robots` group instead of `elastic/observablt-robots-on-call`.

## Why is it important?

There is no rotation anymore for on-call.

### DIFF
--- a/.ci/.bump-go-release-version.yml
+++ b/.ci/.bump-go-release-version.yml
@@ -21,14 +21,14 @@ projects:
       - main
     enabled: true
     labels: dependency
-    reviewer: elastic/observablt-robots
+    reviewer: elastic/observablt-ci
   - repo: apm-integration-testing
     script: .ci/bump-go-release-version.sh
     branches:
       - main
     enabled: true
     labels: dependency
-    reviewer: elastic/observablt-robots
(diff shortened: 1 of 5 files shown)


v1v (commented):
I'd say to use the new team `elastic/observablt-ci` instead, since `elastic/observablt-robots` is the generic one.

amannocci:
> I'd say to use the new team `elastic/observablt-ci` instead, since `elastic/observablt-robots` is the generic one.

Do we maintain all those scripts?

v1v:
> Do we maintain all those scripts?

I don't understand your question, do you mean if we are the ones maintaining this repository?

amannocci:
Yes, you are right. Do we maintain all of them? If that's the case, then we can use the more specific one.

v1v:
> Yes, you are right. Do we maintain all of them? If that's the case, then we can use the more specific one.

I'd say yes, there are just a few files related to the OTEL integration for `pytest` that @kuisathaverat maintains and might eventually move to somewhere else. Since this project is quite specific for the CI, I'd say to use the specific team

v1v (approved):

PR #1834, opened 2022-08-08
feat: enable dependabot for pytest-otel

MoshiBin:
## What does this PR do?

Configures dependabot to track pytest-otel.

## Why is it important?

pytest-otel currently requires opentelemetry-python packages at version 1.11.0; however, 1.11.1 is available. By using dependabot we can keep the OTel deps updated, and catch any breaking changes.

## Related issues

None


kuisathaverat:
ooops!! I did not see this one, I just added it

PR #1835, opened 2022-08-08
Freeze Go version at 1.18.

cmacknz:
Prevents versions bumps to Go 1.19 until we are ready to update, we are
still working to support Go 1.18.


amannocci (approved):
LGTM

PR #1845, opened 2022-08-18
Retry only on transient errors

amannocci:

## What does this PR do?

* Replace the bash retry strategy with the curl retry strategy.
* It will only retry on transient errors (HTTP 408, 429, 500, 502, 503, or 504 response code).

## Why is it important?

* We retry curl requests even if it's not necessary when we build reports.

```txt
15:50:42  INFO: fetchAndPrepareTestCoverageReport (see tests-coverage.json)
15:50:42  INFO: curl <url> -o tests-coverage.json
15:50:42  Retry 1/3 exited 22, retrying in 1 seconds...
15:50:43  Retry 2/3 exited 22, retrying in 2 seconds...
15:50:45  Retry 3/3 exited 22, no more retries left.
15:50:45  jq: error: Could not open file tests-coverage.json: No such file or directory
```


v1v (approved):
Thanks! 
I assume `--retry` is a well available flag in curl, just in case the CI controllers are running a very old OS version

amannocci:
> Thanks! I assume `--retry` is a well-available flag in curl, just in case the CI controllers are running a very old OS version

That's a good question...
It seems to be available since 7.24.0 - January 24 2012 if we refer to https://curl.se/changes.html#7_24_0

v1v:
> pre-commit / pre-commit (pull_request) Failing after 11s

It will fail always as the docker images are stored in the elastic docker registry. Hence this check should be moved back to Jenkins. 

Unrelated to this particular issue

amannocci:
> > pre-commit / pre-commit (pull_request) Failing after 11s
> 
> It will fail always as the docker images are stored in the elastic docker registry. Hence this check should be moved back to Jenkins.
> 
> Unrelated to this particular issue

It should be safe to merge.

cachedout (approved):

PR #1850, opened 2022-08-24
Fallback runE2E to main if a feature branch in Beats/ElasticAgent

v1v:
## What does this PR do?

e2e-testing branch strategy follows the same branch strategy defined in `elastic/beats.git` and `elastic/elastic-agent`, aka, main and release branches. Hence feature branches are not supported.

Avoid feature branches by falling back to `main`

## Why is it important?

The existing branch strategy in the `e2e-testing` needs to be aligned with the consumers one. otherwise things won't work.

This is an alternative and might require some clarification with the teams about what to do and ownership.

Using this approach will make things work but there is a chance the target branch will not contain the expected implementation, if so, what do we do?


v1v:
We have finally ended up to use a branch within the `e2e-testing`

PR #1863, opened 2022-08-31
dependabot: less often

v1v:
## What does this PR do?

Run less often the dependabot bumps

## Why is it important?

There is no need to be up-to-date immediately, so we can reduce the code review overhead/notifications a bit


amannocci (changes requested):

cachedout (approved):
Good change 👍

amannocci (approved):

PR #1865, opened 2022-09-06
Validate yq is installed on beats-ci

amannocci:

## What does this PR do?

Check `yq` is installed on beats-ci.

## Why is it important?

`yq` is needed for the `ingest-dev` project.

### DIFF
--- a/test-infra/beats-ci/test_beats_installed_tools.py
+++ b/test-infra/beats-ci/test_beats_installed_tools.py
@@ -72,3 +72,7 @@ def test_tar_installed(host):
 def test_vault_installed(host):
   cmd = host.run("vault --version")
   assert cmd.rc == 0, "it is required for all the Beats projects"
+
+def test_yq_is_installed(host):
+  cmd = host.run("yq --version")
+  assert cmd.rc == 0, "it is required for ingest-dev project"


test-infra/beats-ci/test_beats_installed_tools.py
```diff
@@ -72,3 +72,7 @@ def test_tar_installed(host):
 def test_vault_installed(host):
   cmd = host.run("vault --version")
   assert cmd.rc == 0, "it is required for all the Beats projects"
+
+def test_yq_is_installed(host):
```
v1v, line 76: This test will run for all the existing workers defined in https://github.com/elastic/apm-pipeline-library/blob/d27475280c35162232d4ebbd2ecc4c1fe6dc7276/.ci/validateWorkersBeatsCi.groovy#L64-L80


and the call to this test is defined in:
- https://github.com/elastic/apm-pipeline-library/blob/main/resources/scripts/jenkins/beats-ci/test.sh#L35
- https://github.com/elastic/apm-pipeline-library/blob/d27475280c35162232d4ebbd2ecc4c1fe6dc7276/resources/scripts/jenkins/beats-ci/test-mac.sh#L35

Will `yq` be available in `MacOS`, `arm` or the static linux workers?

amannocci: It should be available on arm/amd64 ephemeral Linux workers for now.
In that case, could we exclude every unsupported platform as done in some cases?

v1v: The only one that could cause some issues could be the one regarding the `static workers`. Maybe this could be solved by installing `yq` in those static workers, what do you think?

The one for `macos` you can use something like:
- https://github.com/elastic/apm-pipeline-library/blob/main/test-infra/apm-ci/test_installed_tools.py#L44

v1v:
@amannocci , as we are moving away from Jenkins, what do you think if we close this PR?

amannocci:
> @amannocci , as we are moving away from Jenkins, what do you think if we close this PR?

We can close it 👍

PR #1866, opened 2022-09-06
ci: pre-commit in Jenkins

v1v:
## What does this PR do?

Run `pre-commit` in Jenkins.
Provision one CI worker and use the same for the different sequential stages, this will help with faster builds.

## Why is it important?

It requires to access the internal docker registry, hence we don't want to complicate the GitHub actions with some credentials


amannocci (approved):
LGTM.
PS: Maybe we will need to delete the rule `pre-commit` since we now have `Check pre-commit`

PR #1882, opened 2022-09-15
fix: path filter incorrect

kuisathaverat:
## What does this PR do?

* it fixes the filter path 
* it reports JUnit test results

### DIFF
--- a/.github/workflows/pytest_otel.yml
+++ b/.github/workflows/pytest_otel.yml
@@ -8,6 +8,7 @@ on:
       - 'resources/scripts/pytest_otel/**'
       - '.github/workflows/pytest_otel.yml'
   push:
+    branches: main
     paths:
       - 'resources/scripts/pytest_otel/**'
 
--- a/.github/workflows/pytest_otel.yml
+++ b/.github/workflows/pytest_otel.yml
@@ -32,8 +32,3 @@ jobs:
         run: make -C resources/scripts/pytest_otel test
       - name: it-test
         run: make -C resources/scripts/pytest_otel it-test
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest_otel.yml
```diff
@@ -31,3 +31,8 @@ jobs:
         run: make -C resources/scripts/pytest_otel test
       - name: it-test
         run: make -C resources/scripts/pytest_otel it-test
+      - uses: rainbow-duck-games/unity-test-publisher@82614fdbce4f519c5f75b24511d94250af6c04d9
+        if: ${{ always() }} # Avoid skipping on failed tests
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          reportPaths: resources/scripts/pytest_otel/**/junit-*.xml
```
v1v, line 38: no strong opinion, though I also researched other publishers such as:
- EnricoMi/publish-unit-test-result-action@f00bb47e1d89c6d98071d1b69e16a2b63c7fbfc1
- dorny/test-reporter@v1

as they suggested to use another event instead running within the default workflow, as per https://github.com/EnricoMi/publish-unit-test-result-action#support-fork-repositories-and-dependabot-branches

.github/workflows/pytest_otel.yml
```diff
@@ -5,10 +5,10 @@ on:
   workflow_dispatch:
   pull_request:
     paths:
-      - '/resources/scripts/pytest_otel/**'
+      - 'resources/scripts/pytest_otel/**'
```
v1v, line 8: Can you also add this workflow too?

```suggestion
      - 'resources/scripts/pytest_otel/**'
      - '.github/workflows/pytest_otel.yml'
```

v1v (approved):

PR #1924, opened 2022-10-24
feat: GitHub Actions

kuisathaverat:
## What does this PR do?

More GitHub Actions

* feat: setup vault CLI
* feta: setup git


pazone (approved):

.github/actions/pre-commit/action.yml
```diff
@@ -0,0 +1,24 @@
...
+        fetch-depth: 0
+    - uses: actions/setup-python@v4
+      with:
+        python-version: 3.9
+
+    - name: Run pre-commit on PRs
+      uses: pre-commit/action@v3.0.0
+      if: ${{ github.event.pull_request }}
+      with:
+        extra_args: --from-ref ${{ github.event.pull_request.base.sha }} --to-ref ${{ github.event.pull_request.head.sha }}
```
v1v, line 20: neat!!

.github/actions/setup-git/action.yml
```diff
@@ -0,0 +1,37 @@
...
+      shell: bash
+      run: |
+        git config --global user.name "${GIT_USER}"
+        git config --global user.email "${GIT_EMAIL}"
+        gh auth setup-git
+      env:
+        GIT_USER: "${{ inputs.username }}"
+        GIT_EMAIL: "${{ inputs.email }}"
+        GIT_TRACE: "${{ inputs.trace }}"
+        GITHUB_TOKEN: ${{ inputs.token }}
```
v1v, line 37: I guess, this will never expose the password in the logs, won't it?

v1v (approved):
just a comment regarding the token, other than that LGTM

.github/actions/setup-git/action.yml
kuisathaverat: it delegates the responsibility to mask the secret to the generator of the GITHUB_TOKEN, I mean is you pass `${{ secrets.GITHUB_TOKEN }}` of something you read from the Vault, or a Token got with https://github.com/elastic/apm-pipeline-library/blob/main/.github/actions/github-token/action.yml the token is masked in the action that gets the token. There is no action that masks tokens that are received as an input they must be already masked.

PR #1946, opened 2022-11-09
action: support go-json

v1v:
## What does this PR do?

Add support for the `go-json` reporter

## Why is it important?

Simplify consumers so they don't need to do any tricks regarding the transformation

## Follow up

Support this process in the upstream project (see https://github.com/dorny/test-reporter/issues/189)

## Issues

Required by https://github.com/elastic/ecs-logging-go-zap/pull/49


.github/actions/test-report/action.yml
```diff
@@ -59,8 +60,41 @@ inputs:
...
+      with:
+        name: ${{ inputs.name }}
+        path: "go-json-test-results.xml"
+        reporter: 'java-junit'
+        list-suites: ${{ inputs.list-suites }}
+        list-tests: ${{ inputs.list-tests }}
+        fail-on-error: ${{ inputs.fail-on-error }}
+        only-summary: ${{ inputs.only-summary }}
+
+    # If no go-json reporter then as usual
```
kuisathaverat, line 94: sounds like a new different action, it uses parameters to choose one action or another, makes things complex, and can cause side effects on projects that do not use Go.

v1v: I'll refactor this in a different action, so the current action will call the other one then

kuisathaverat: I understand that the goal is to use always the same pattern in all projects, it is fine, in that case I preferred to have the actions `.github/actions/test-report-go`, `.github/actions/test-report-base`, and one to join both `.github/actions/test-report` that has only the conditionals

v1v: See now, I'll contribute to the upstream project, and likely we will move away from this implementation.

.github/actions/test-report/action.yml
```diff
@@ -59,7 +60,19 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Report go test results
+      if: contains('go-json', inputs.reporter)
+      uses: elastic/apm-pipeline-library/.github/actions/test-go-report@feature/go-json-reporter
```
kuisathaverat, line 65: IIRC to pass an expression to `uses`, if so when we have more than two alternatives we probably want to change to have a step that selects the action and puts the name in an output we use here.

.github/actions/test-report/action.yml
```diff
@@ -59,7 +60,19 @@ inputs:
...
+      with:
+        name: ${{ inputs.name }}
+        path: ${{ inputs.path }}
+        list-suites: ${{ inputs.list-suites }}
+        list-tests: ${{ inputs.list-tests }}
+        fail-on-error: ${{ inputs.fail-on-error }}
+        only-summary: ${{ inputs.only-summary }}
```
kuisathaverat, line 72: these inputs are only used in this report, I am not sure if it is better to use an `options` input and define the extra parameter in YAML to JSON in that field.

```yaml
jobs:
  process:
    runs-on: ubuntu-latest
    steps:
...
      - uses: elastic/apm-pipeline-library/.github/actions/test-report@main
        name: foo
        path: /bar
        options: |
          op01: val01
          op02: var02
```

kuisathaverat: with JSON the transformation is direct https://docs.github.com/en/actions/learn-github-actions/expressions#tojson with YAML we have to implement a simple action to read a YAML input-key

kuisathaverat: or transform YAML to JSON 
```bash
OPTIONS=${{ input.options }}
JSON=$(echo "${OPTIONS}" | yq -c .) 
echo "json=${JSON}" >> $GITHUB_OUTPUT 
```

v1v:
I'll close this PR as stalled , and likely to contribute to the upstream instead

PR #1959, opened 2022-11-24
Remove spurious quote when looking for .go-version

jsoriano:
## What does this PR do?

Remove a spurious quote that is producing an unexpected string.

## Why is it important?

This prevents `goDefaultVersion` to find the Go version in  `env.BASE_DIR`, as it tries to find `.go-version'`.

## Related issues

Found in https://github.com/elastic/integrations/pull/4717, where the default version of Go is being unexpectedly installed.


v1v (approved):
oh dear! I somehow got crazy in the past for a similar issue and I could not find the typo! Thanks so much Jaime

PR #1974, opened 2022-12-14
Add support for snapshoty github action

amannocci:

## What does this PR do?

* Add support for snapshoty github action.

## Why is it important?

* Allow us to use snapshoty in github action.


.github/actions/snapshoty/dist/index.js
```diff
@@ -0,0 +1,4160 @@
+require('./sourcemap-register.js');/******/ (() => { // webpackBootstrap
```
reakaleek, line 1: Can this file be minified?

.github/actions/snapshoty/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    required: true
+  gcsPrivateKeyId:
+    description: 'Google Cloud private key id of the service account'
+    required: true
+  gcsProject:
+    description: 'Google Cloud project id of the service account'
+    required: true
+runs:
+  using: 'node16'
+  main: 'dist/index.js'
```
kuisathaverat, line 24: I think that a composite with a run step launching the command suits better to the use case

kuisathaverat (approved):

reakaleek (approved):

PR #1981, opened 2022-12-23
Fix docker-layer-caching action inputs

reakaleek:
## Related issues
Closes #ISSUE


reakaleek:
Tried to reproduce the error happening in https://github.com/elastic/ecs-logging-ruby/actions/runs/3765830128/jobs/6402632859.

But it seems like it only happens for 1 matrix item.

Closing

PR #1985, opened 2022-12-27
Add flattenMap

mrodm:
## What does this PR do?

Add a new function `flattenMap` to flatten the map given as a parameter. By default it uses the dot as separator, but it can be customized.

Example:

```groovy
def m = ["a": ["b": ["c":[1,2,3], "d":[4,5,6] ] ] ]

def flattened = flattenMap(map: m)

// flattened == ["a.b.c": [1,2,3], "a.b.d": [4,5,6] ]
```

## Why is it important?

For those maps that nested fields can be defined:
- using dots in the key:
  ```yaml
  conditions:
    kibana.version: "^8.0.0"
  ```
- with nested maps:
  ```yaml
  conditions:
    kibana:
      version: "^8.0.0"   
  ```

This function (`flattenMap`) allows the developers to have a common representation with all the fields flattened. In the example given above, the resulting map would be:
```yaml
conditions.kibana.version: "8.0.0"
```

Example of current usage: https://github.com/elastic/integrations/blob/main/.ci/Jenkinsfile#L304-L315


v1v (approved):

jsoriano (approved):
:+1:

v1v:
@mrodm, I just triggered a new release, so this new step will be available in a few minutes. This [build](https://apm-ci.elastic.co/job/apm-shared/job/apm-pipeline-library-mbp/job/main/578/) is the one in charge of the release.

mrodm:
Thanks @v1v !

PR #1986, opened 2022-12-28
Migrate build-docker-images to GH Actions Workflow

reakaleek:
## What does this PR do?

- Migrate build-docker-images to GH Actions Workflow.
- Add json schema to be able to validate the config
  - Is added to pre-commit validation
  - Can also be added to the IDE 

## Why is it important?

Jenkins 2 GH Actions Migration

## To Do
- [x] Migration Jenkinsfile to GH Actions Workflow
- [x] Delete Job DSL Configs
- [x] Add Documentation
- [ ] Delete Jobs?
- [ ] Update sheet

## Follow Up
- https://github.com/elastic/apm-pipeline-library/issues/1992

PR #1997, opened 2023-01-04
Fix publish-docker-images workflow v2

reakaleek:
## What does this PR do?

The fix in https://github.com/elastic/apm-pipeline-library/pull/1995 did not work as expected.

The culprit is that in GH Actions Expressions, they do loose comparisons, which causes unexpected behaviour:

```javascript
${{ '' != false }} // false
${{ '' == false }} // true
${{ null != false }} // false
${{ null == false }} // true

// when used to evaluate matrix.push, which can be true, false or undefined
// ${{ matrix.push != false }}
```

By putting the `matrix.push` into a environment variable, the expression will be evaluated using string comparison and therefore gives the expected result.

## Why is it important?

The publish-docker-images workflow is already merged and the old build-docker-images Jenkinsfile is already removed. Hence, it's not working ATM.

## Related issues
- https://github.com/elastic/apm-pipeline-library/pull/1995
- https://github.com/elastic/apm-pipeline-library/pull/1986

### DIFF
--- a/.github/workflows/publish-docker-images.yml
+++ b/.github/workflows/publish-docker-images.yml
@@ -74,6 +74,7 @@ jobs:
         run: |
           echo "NAME=${{ matrix.name }}" >> $GITHUB_ENV
           echo "TAG=${{ matrix.tag || 'latest' }}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


kuisathaverat:
It deserves a [WAT](https://www.destroyallsoftware.com/talks/wat)

PR #1999, opened 2023-01-05
Fix working directory in publish-docker-images workflow

reakaleek:
## What does this PR do?

Sets the working directory for each step in publish-docker-images workflow using the `defaults` attribute.

## Why is it important?

I forgot to add the working directory for the push step and the workflow is failing for images with a custom `push_script`.


reakaleek:
Sorry, @kuisathaverat 😅 another fix is necesary.

But I'm confident this is the last one.

kuisathaverat (approved):

PR #2000, opened 2023-01-09
fix: use observability Vault namespace

kuisathaverat:
## What does this PR do?

It changes to use the `observability` namespace instead the `apm` namespace for grabbing vault secrets.

## Why is it important?

We have to update only the `observability` secrets. The use of the `apm` should be deprecated.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -141,7 +141,7 @@ pipeline {
           steps {
             deleteDir()
             unstash 'source'
-            dockerLogin(secret: "secret/apm-team/ci/docker-registry/prod",
+            dockerLogin(secret: "secret/observability-team/ci/docker-registry/prod",
               registry: "docker.elastic.co"
             )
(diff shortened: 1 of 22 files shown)


reakaleek (approved):

v1v (approved):

v1v:
IIRC, the UTs are coupled with the Vault secret path, so https://github.com/elastic/apm-pipeline-library/blob/0ef363d6daa295e4dc3a70d769eaa5ab1dad57e7/src/test/groovy/ApmBasePipelineTest.groovy#L40-L64 might need to be changed accordingly

kuisathaverat:
@v1v there are two secrets that are only on the `apm` namespace `secret/apm-team/ci/apm-agent-dotnet-azure` and `secret/apm-team/ci/elastic-observability-npmjs` due to I dunno the implications of moving those secrets, I have only moved the ones that are duplicated.

PR #2007, opened 2023-01-16
action: export traces with the OTEL GH action

v1v:
use the original action rather than the forked one and enable to all the workflows


reakaleek (approved):

v1v:
> 🤔 to avoid having to create OTEL secrets all over the place, I think we should use the vault.

That's a good point. I had in mind to work on this in a follow up, but let me try to put something as I need to roll out this almost in every Oblt project...

PR #2008, opened 2023-01-16
action: use explicit workflow names

v1v:
* rename workflows with some IDs rather than meaningful names
* user explicit workflow names otherwise the workflow won't run as it's required

### DIFF
--- a/.github/workflows/actions-test.yml
+++ b/.github/workflows/actions-test.yml
@@ -1,6 +1,6 @@
 ---
 ## Test custom actions in .github/actions/
-name: actions
+name: actions-test
 
 on:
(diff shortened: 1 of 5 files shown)


reakaleek (approved):
I had this in my mind recently:

Maybe we can introduce something like an elastic observability GH actions naming convention and maybe even a linter for it?

v1v:
> Maybe we can introduce something like an elastic observability GH actions naming convention and maybe even a linter for it. WDYT?

That's a great idea! Let me poke you in a private repository that contains some ideas about reusable actions

v1v:
Much better now


image

PR #2011, opened 2023-01-16
action: composite cannot access to secrets

v1v:
## What does this PR do?

Restore the secrets since:

## Why is it important?

> The secrets context contains the names and values of secrets that are available to a workflow run. The secrets context is not available for composite actions due to security reasons.

See https://docs.github.com/en/actions/learn-github-actions/contexts#secrets-context


reakaleek:
Unfortunate

reakaleek (approved):

v1v:
> Unfortunate, but thinking about it.. it makes sense

My bad, I somehow decided to go for it, rather than reading the manual :/ 

But I'll remember it this for the future, thanks for making the point avoid default values! 🙏

PR #2030, opened 2023-01-24
actions: use hashicorp/vault-action to prepare the github release

v1v:
## What does this PR do?

Set a new env variable to be populated for the releases.

## Why is it important?

Somehow it is not working with the PAT

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
           email: ${{ env.GIT_EMAIL }}
           token: ${{ env.GITHUB_TOKEN }}
 
-      - name: configure git
+      - name: Configure git
         run: |
           git remote add origin-1 "${REMOTE_URL}"
           git fetch --force --tags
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -29,39 +29,42 @@ jobs:
...
         run: |
           git remote add origin-1 "${REMOTE_URL}"
           git fetch --force --tags
           git checkout main
         env:
           REMOTE_URL: ${{ github.event.repository.html_url }}
 
-      - name: Cache local Maven repository
-        uses: actions/cache@v3
+      - name: Set up JDK 11
```
v1v, line 55: Similarly done in the CI pipeline, so using the same approach with caching

.github/workflows/release.yml
```diff
@@ -29,39 +29,42 @@ jobs:
...
-            ${{ runner.os }}-maven-
+          java-version: '11'
+          distribution: 'adopt'
+          cache: 'maven'
 
-      - name: mvn release
+      - name: Mvn release
         run: ./mvnw -B -V release:prepare release:perform --batch-mode -Darguments="-DskipTests=true --batch-mode"
+        env:
+          GITHUB_TOKEN: ${{ env.GITHUB_TOKEN_PAT }}
```
v1v, line 65: Reset the variable that was read in the vault secret

PR #2057, opened 2023-02-06
ci: remove jobdsl, ci metrics and test Beats CI

v1v:
## What does this PR do?

Remove jobs creation with JobDSL and CI metrics gathering ot test Beats Infra

## Why is it important?

Our current Jenkins ecosystem is deprecated.


amannocci (approved):
LGTM

PR #2062, opened 2023-02-07
shellcheck: use official pinned version 

v1v:
## What does this PR do?

Use the official docker pinned version with the latest `0.9.0` version

## Why is it important?

Fix the issue while compiling the project from the source code. I tried by bumping the OS version and went through the build system but somehow I could not make it work.

Since we don't use this often, I'd say to use this approach in favour of having something working. Otherwise we could remove it from the project since I don't see much benefit for buidlilng over and over from scratch for the same pinned version

### DIFF
--- a/.ci/docker/shellcheck/Dockerfile
+++ b/.ci/docker/shellcheck/Dockerfile
@@ -1,5 +1,5 @@
 # Build-only image
-FROM ubuntu:18.04 AS build
+FROM ubuntu:20.04 AS build
 USER root
 WORKDIR /opt/shellCheck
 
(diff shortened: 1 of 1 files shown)


v1v:
though there are further errors with the linking:

```
#12 123.3 [18 of 19] Compiling ShellCheck.Checker ( src/ShellCheck/Checker.hs, src/ShellCheck/Checker.o )
#12 123.5 [19 of 19] Compiling Main             ( shellcheck.hs, shellcheck.o )
#12 123.9 Linking shellcheck ...
#12 125.2 /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libpthread.a(pthread_mutex_init.o)(.note.stapsdt+0x14): error: relocation refers to local symbol "" [2], which is defined in a discarded section
#12 125.2   section group signature: "(null)"
```

reakaleek (approved):

PR #2085, opened 2023-02-15
action: bump-aliases use docker login

v1v:
## What does this PR do?

`bump-aliases` uses the environmental context that has already access to the prviate docker registry

## Why is it important?

Use docker-login and env variables are reseted after the login has happened, hence it's not accessible by env variables

## Test

I used the `.ci/bump-aliases.yml` locally with the username/password as I was already logged in and it worked.


amannocci (approved):
LGTM

PR #2087, opened 2023-02-16
action: is-member-elastic-org

v1v:
## What does this PR do?

Support action to verify if the given GitHub user is member of the `elastic` org

## Why is it important?

This will help us to move forward with the GitHub command ops to run specific actions based on specific GitHub commands

## Test

see [this workflow](https://github.com/elastic/apm-pipeline-library/actions/runs/4195806309/jobs/7275877301) that tested these changes

## Issues

https://github.com/elastic/apm-pipeline-library/pull/2089 is another alternative using the `gh cli` with a composite action. Which one do you prefer?


v1v:
I'll close this implementation for now, as the one in https://github.com/elastic/apm-pipeline-library/pull/2089 seems to work nicely by using the CLI and less code to maintain

PR #2094, opened 2023-02-20
action: workflow event composite action

v1v:
## What does this PR do?

GitHub composite action to trigger workflows between GitHub repositories using the [repository_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch)

## Why is it important?

This will help us to interconnect different workflows with different even types, such as, a GitHub command will trigger a workflow in a different repository (even private repositories)

## Tasks

- [x] Define the payload format
- [x] Test this out

## Test

This [build](https://github.com/elastic/apm-pipeline-library/actions/runs/4224575367/jobs/7335706206), caused by  https://github.com/elastic/apm-pipeline-library/pull/2094/commits/0c4069e90f643ffaf45a397894024f0ddb19267a, triggered the build in one of our internal playground GitHub repositories:

image

### DIFF
--- a/.github/workflows/test-workflow-run.yml
+++ b/.github/workflows/test-workflow-run.yml
@@ -26,4 +26,4 @@ jobs:
           repository: "elastic/observability-robots-playground"
           type: my-event
           token: ${{ env.GITHUB_TOKEN }}
-          payload: '{ "sha": "${{ github.event.pull_request.head.sha }}", repository: "${{ github.repository }}" }'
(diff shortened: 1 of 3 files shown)


.github/workflows/test-workflow-run.yml
```diff
@@ -0,0 +1,29 @@
+---
+name: test-workflow-run
+
+on:
+  pull_request:
```
v1v, line 5:
```suggestion
```

reakaleek (approved):

PR #2096, opened 2023-02-21
action: use GH_TOKEN

v1v:
## What does this PR do?

Use GH_TOKEN for the gh cli

## Why is it important?

Avoid header but using env variables for simplicity

### DIFF
--- a/.github/actions/is-member-elastic-org/action.yml
+++ b/.github/actions/is-member-elastic-org/action.yml
@@ -18,11 +18,12 @@ runs:
       name: Check if user is member of the Elastic org
       run: |
         if gh api \
-          -H "Authorization: Bearer ${{ inputs.token }}" \
           -H "Accept: application/vnd.github+json" \
           /orgs/elastic/members/${{ inputs.username }} ; then
           echo "result=true" >> $GITHUB_OUTPUT
         else
           echo "result=false" >> $GITHUB_OUTPUT
         fi
+      env:
+        GH_TOKEN: ${{ inputs.token }}
       shell: bash
--- a/.github/actions/comment-reaction/action.yml
+++ b/.github/actions/comment-reaction/action.yml
@@ -21,7 +21,8 @@ runs:
         gh api \
           --method POST \
           -H "Accept: application/vnd.github+json" \
-          -H "Authorization: Bearer ${{ inputs.token }}" \
           /repos/${{ inputs.repository }}/issues/comments/${{ inputs.commentId }}/reactions \
           -f content='${{ inputs.emoji }}'
+      env:
+        GH_TOKEN: ${{ inputs.token }}
       shell: bash


.github/actions/is-member-elastic-org/action.yml
```diff
@@ -18,11 +18,12 @@ runs:
...
         if gh api \
-          -H "Authorization: Bearer ${{ inputs.token }}" \
           -H "Accept: application/vnd.github+json" \
           /orgs/elastic/members/${{ inputs.username }} ; then
           echo "result=true" >> $GITHUB_OUTPUT
         else
           echo "result=false" >> $GITHUB_OUTPUT
         fi
+      env:
+        GH_TOKEN: ${{ inputs.token }}
```
reakaleek, line 28: I think you can access `${{ github.token }}` from an action.

reakaleek: this is how we do it in the `setup-git` action https://github.com/elastic/apm-pipeline-library/blob/05b38c3302568a0d040e38d988ce01f41a48cf11/.github/actions/setup-git/action.yml#L23

v1v: Reusable workflows has secrets but composite actions are limited to inputs, so it won't work

reakaleek: Yes, but technically `${{ github.xyz }}` is not a secret.

reakaleek: I think it's worth to verify as it would reduce 1 input variable to think about.

v1v: https://github.com/elastic/apm-pipeline-library/actions/runs/4232854415/jobs/7353092719 is the one verifying this behaviour, but the default token does not have enough permissions to access the GH API

https://github.com/elastic/apm-pipeline-library/pull/2098/commits/ac7a51ca45a4cd18fd9cd9d88a87a274e3972ac1 is the specific commit on the top of this PR that verified it cannot be used the defaulted token

reakaleek: ok 😅 thank you very much for checking.

but from what I understand, generally it works.. but it doesn't work for this usecase?

reakaleek (approved):

.github/actions/is-member-elastic-org/action.yml
v1v: Yep, it works but in this case the gh api call requires different permissions, since it queries the `org` for a particular user, while the defautl ephemeral GitHub token does not have permissions to access that particular API, or I could not find a way to configure the scope to allow those permissions in https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

reakaleek: I see.. yeah, I also kind of checked.. but from a very superficial check.. it doesn't seem like we can grant those org read permissions.. and a PAT is necessary.

v1v: Yep, that's my feeling too

PR #2097, opened 2023-02-21
action: oblt-cli action

v1v:
## What does this PR do?

Composite action to interact with the oblt-cli tool

## Why is it important?

Simplify consumers so they can directly use the CLI but without the need to download the binary

## Test

See this [build](https://github.com/elastic/apm-pipeline-library/actions/runs/4232636127/jobs/7352610773) that triggered the `oblt-cli`

PR #2099, opened 2023-02-21
action:  oblt-cli-create-ccs composite action

v1v:
## What does this PR do?

oblt-cli-create-ccs composite action to create a CCS cluster

## Why is it important?

Specialise composite actions so consumers don't need to know about the oblt-cli internals.

## Related issues

Requires https://github.com/elastic/apm-pipeline-library/pull/2097

### DIFF
--- a/.github/actions/oblt-cli-create-ccs/README.md
+++ b/.github/actions/oblt-cli-create-ccs/README.md
@@ -46,3 +46,4 @@ Following inputs can be used as `step.with` keys
 | `kibana-docker-image`       | String  | Optional                    | Force to use a Docker image for the Kibana Deployment |
 | `slackChannel`              | String  | `#observablt-bots`          | The slack channel to be configured in the oblt-cli. |
 | `token`                     | String  | Mandatory                   | The GitHub token with permissions fetch releases. |
+| `username`                  | String  | `apmmachine`                | Username to show in the deployments with oblt-cli, format: [a-z0-9]. |
--- a/.github/actions/oblt-cli-create-ccs/README.md
+++ b/.github/actions/oblt-cli-create-ccs/README.md
@@ -1,0 +1,48 @@
+## About
+
+GitHub Action to run the oblt-cli wrapper to create a CCS cluster
+
+* [Usage](#usage)
+  * [Configuration](#configuration)
+* [Customizing](#customizing)
+  * [inputs](#inputs)
+
+## Usage
+
+### Configuration
+
+Given the CI GitHub action:
+
+```yaml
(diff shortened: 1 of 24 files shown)


.github/actions/oblt-cli-create-ccs/action.yml
```diff
@@ -0,0 +1,55 @@
+name: 'Oblt-cli create ccs'
+description: 'Run the oblt-cli wrapper to create a CCS cluster.'
+inputs:
+  remote-cluster:
```
v1v, line 4: I decided to use this input variable name with `-` to match the oblt-cli commands. 

We could revert to be camelcase as we do with other actions, but somehow I thought it might be better to use the same input format for those parameters matching the oblt-cli flags

.github/actions/oblt-cli-create-ccs/README.md
```diff
@@ -0,0 +1,49 @@
...
+Following inputs can be used as `step.with` keys
+
+| Name                        | Type    | Default                     | Description                        |
+|-----------------------------|---------|-----------------------------|------------------------------------|
+| `remote-cluster`            | String  | Mandatory                   | The Oblt cluster to use. |
+| `cluster-name-prefix`       | String  | Optional                    | Prefix to be prepended to the randomised cluster name |
+| `cluster-name-suffix`       | String  | Optional                    | Suffix to be appended to the randomised cluster name |
+| `apm-docker-image`          | String  | Optional                    | Force a Docker image for the APM Deployment |
+| `elasticsearch-docker-image`| String  | Optional                    | Force to use a Docker image for the Elasticserach Deployment |
+| `kibana-docker-image`       | String  | Optional                    | Force to use a Docker image for the Kibana Deployment |
```
v1v, line 46: @kuisathaverat , IIUC, `elasticsearch-docker-image` is the only one available for `create ccs` even `apm` or `kibana` are available through flags, they are not consumed in the templating, as far as I see.

If that's the case, I guess these inputs are not required.

kuisathaverat: for CCS Kibana and APm are disabled so those parameters does not have sense

kuisathaverat (approved):

.github/workflows/test-oblt-cli-create-ccs.yml
```diff
@@ -0,0 +1,28 @@
+---
+name: test-oblt-cli-create-ccs
+
+on:
+  pull_request:
```
v1v, line 5:
```suggestion
```

PR #2125, opened 2023-03-06
Publish jruby images in dockerhub

reakaleek:
## What does this PR do?

- Adds the ability to push to dockerhub
- Pushes jruby images to dockerhub instead

## Why is it important?

This is needed to be able to run PR tests in https://github.com/elastic/apm-agent-ruby/pull/1366 without the necessity of GH secrets.

## Related issues
- Depends on https://github.com/elastic/apm-agent-ruby/pull/1367
- Needed in https://github.com/elastic/apm-agent-ruby/pull/1366

### DIFF
--- a/.ci/.docker-images.yml
+++ b/.ci/.docker-images.yml
@@ -279,13 +279,13 @@ images:
     build_opts: "--build-arg NODE_VERSION=8.6"
 
   # APM Agent Ruby Docker images
-
+  # dockerhub
   - <<: *apm-agent-ruby-docker-images
     name: "apm-agent-jruby"
     working_directory: ".ci/docker/jruby"
-    build_script: "./run.sh --action build --registry ${REGISTRY}/${PREFIX}"
-    test_script: "./run.sh --action test --registry ${REGISTRY}/${PREFIX}"
-    push_script: "./run.sh --action push --registry ${REGISTRY}/${PREFIX}"
+    build_script: "./run.sh --action build --registry 'elasticobservability'"
+    test_script: "./run.sh --action test --registry 'elasticobservability'"
+    push_script: "./run.sh --action push --registry 'elasticobservability'"
 
   - <<: *apm-agent-ruby-docker-images
     tag: "ruby-3.0"
--- a/.github/workflows/publish-docker-images.yml
+++ b/.github/workflows/publish-docker-images.yml
@@ -73,6 +73,14 @@ jobs:
           roleId: ${{ secrets.VAULT_ROLE_ID }}
           secretId: ${{ secrets.VAULT_SECRET_ID }}
 
+      - uses: elastic/apm-pipeline-library/.github/actions/docker-login@current
+        with:
+          registry: ""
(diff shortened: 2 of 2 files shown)


.ci/.docker-images.yml
```diff
@@ -277,15 +277,29 @@ images:
...
+
+  # dockerhub
   - <<: *apm-agent-ruby-docker-images
     name: "apm-agent-jruby"
     working_directory: ".ci/docker/jruby"
-    build_script: "./run.sh --action build --registry ${REGISTRY}/${PREFIX}"
-    test_script: "./run.sh --action test --registry ${REGISTRY}/${PREFIX}"
-    push_script: "./run.sh --action push --registry ${REGISTRY}/${PREFIX}"
+    build_script: |
+      EXCLUDE=7-jdk ./run.sh --action build --registry 'elasticobservability'
```
v1v, line 295: I know this is a draft PR, but I wonder whether this automation should not know anything about what versions should be built, but delegate that responsibility to the `run.sh`, what do you think?

reakaleek: do you mean the `EXCLUDE` part?

In that case, I kind of agree.. I was torn between the 3 options of 
- using `EXCLUDE` here
- removing the directory entirely
- or excluding in the the `run.sh` script

v1v: > do you mean the EXCLUDE part?

Yes - sorry for my vague sentence!

reakaleek: FYI https://github.com/elastic/apm-agent-ruby/pull/1367/commits/1089887c126a47ca31be9e7d313aa120a6b4c846

.github/workflows/publish-docker-images.yml
```diff
@@ -73,6 +73,14 @@ jobs:
           roleId: ${{ secrets.VAULT_ROLE_ID }}
           secretId: ${{ secrets.VAULT_SECRET_ID }}
 
+      - uses: elastic/apm-pipeline-library/.github/actions/docker-login@current
+        with:
+          registry: ""
```
v1v, line 78: what's the reason for being empty?

v1v: https://github.com/elastic/opbeans-node/blob/70da0522a03b365ad44d4d576d74c698e0ee684c/.github/workflows/release.yml#L21-L22 uses the registry explicitly

reakaleek: Ah ok,

I tried `index.docker.io/v1/`  which did not work.

Then I tried `""` because it then uses the default registry, which is dockerhub.

reakaleek: https://github.com/elastic/apm-pipeline-library/pull/2125/commits/037d34baf05dd0130e4e3a77da74dd6e7c38298a

v1v (approved):

reakaleek:
Successful run here: https://github.com/elastic/apm-pipeline-library/actions/runs/4372814248/jobs/7650217387

PR #2126, opened 2023-03-06
Add check-dependent-jobs action

reakaleek:
## What does this PR do?

Adds an action that outputs the combined statuses of the provided needs context

## Why is it important?

Suggestion from https://github.com/elastic/apm-agent-ruby/pull/1366#discussion_r1126480752

can be used in:
- https://github.com/elastic/apm-agent-python/blob/e6b807d3bd97b843da628763722a1d33185d6f44/.github/workflows/test.yml#L142

## Related issues
- https://github.com/elastic/apm-agent-ruby/pull/1366#discussion_r1126480752

### DIFF
--- a/.github/actions/are-dependent-jobs-successful/action.yml
+++ b/.github/actions/are-dependent-jobs-successful/action.yml
@@ -1,6 +1,6 @@
 ---
 
-name: combine-needs-status
+name: are-dependent-jobs-successful
 description: |
   Evaluates the combined the status results of the provided needs context.
 inputs:
--- a/.github/actions/check-dependent-jobs/README.md
+++ b/.github/actions/check-dependent-jobs/README.md
@@ -44,7 +44,7 @@ jobs:
         uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
         with:
           needs: ${{ toJSON(needs) }}
-      - run: ${{ steps.is_success.check.isSuccess }} # should exit with 1 or 0.
+      - run: ${{ steps.check.isSuccess }} # should exit with 1 or 0.
 ```
 
 ## Customizing
--- a/.github/actions/check-dependent-jobs/README.md
+++ b/.github/actions/check-dependent-jobs/README.md
@@ -40,11 +40,11 @@ jobs:
       - job-a
       - job-b
     steps:
-      - id: is_success
+      - id: check
         uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
         with:
           needs: ${{ toJSON(needs) }}
-      - run: ${{ steps.is_success.outputs.isSuccess }} # should exit with 1 or 0.
+      - run: ${{ steps.is_success.check.isSuccess }} # should exit with 1 or 0.
 ```
 
 ## Customizing
--- a/.github/actions/check-dependent-jobs/README.md
+++ b/.github/actions/check-dependent-jobs/README.md
@@ -3,7 +3,7 @@
 Evaluates the combined the status results of the provided needs context.
 This is useful for creating a single status check.
 
-That status check can the be set as required status check or it can be used
+That status check can then be set as required status check, or it can be used
 in combination with the `notify-built-status` action.
 
 * [Usage](#usage)
@@ -41,10 +41,10 @@ jobs:
       - job-b
     steps:
       - id: is_success
-        uses: elastic/apm-pipeline-library/.github/actions/are-dependent-jobs-successful@current
+        uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
         with:
           needs: ${{ toJSON(needs) }}
-      - run: ${{ steps.is_success.outputs.isSuccess }}
+      - run: ${{ steps.is_success.outputs.isSuccess }} # should exit with 1 or 0.
 ```
 
 ## Customizing
--- a/.github/actions/check-dependent-jobs/README.md
+++ b/.github/actions/check-dependent-jobs/README.md
@@ -44,7 +44,7 @@ jobs:
         uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
         with:
           needs: ${{ toJSON(needs) }}
-      - run: ${{ steps.check.isSuccess }} # should exit with 1 or 0.
+      - run: ${{ steps.check.outputs.isSuccess }} # should exit with 1 or 0.
 ```
 
 ## Customizing
--- a/.github/actions/check-dependent-jobs/action.yml
+++ b/.github/actions/check-dependent-jobs/action.yml
@@ -1,6 +1,6 @@
 ---
 
-name: are-dependent-jobs-successful
+name: check-dependent-jobs
 description: |
   Evaluates the combined the status results of the provided needs context.
 inputs:
--- a/.github/actions/combine-needs-status/action.yml
+++ b/.github/actions/combine-needs-status/action.yml
@@ -2,14 +2,20 @@
 
 name: combine-needs-status
 description: |
-  Combines the status results of the provided needs context and exits with 0 if all jobs are successful
-  or with 1 if at least one job is failing.
+  Evaluates the combined the status results of the provided needs context.
 inputs:
   needs:
     required: true
     description: needs context as JSON string
+outputs:
+  isSuccess:
+    description: the evaluated result of all provided jobs in the needs context.
+    value: ${{ steps.test.outputs.isSuccess }}
 runs:
   using: composite
   steps:
-    - run: test $(echo '${{ inputs.needs }}' | jq -s 'map(.[].result) | all(.=="success")') = 'true'
+    - id: test
+      run: |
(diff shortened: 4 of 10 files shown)


.github/actions/combine-needs-status/action.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+name: combine-needs-status
```
reakaleek, line 3: Open for different naming suggestions @elastic/observablt-ci

v1v: What do you think this other approach?

1) composite action will return a result (boolean)
2) Name of the composite actions will something like `are-dependent-jobs-success`?

IIUC, the current implementation will return errorlevel 1 if no matches with the `test` command.

I somehow feel we can use a boolean output to the composite action, so consumers can use the if condition like:


```

Comment (reakaleek) [reply .github/actions/combine-needs-status/action.yml:3]: I like the idea.

just writing this here to let it sink.

there is also the usecase for setting the "single status check".
this would become something like:

```yaml
  - id: is_success
    uses: elastic/apm-pipeline-library/.github/actions/are-dependent-jobs-success@current
    with: ${{ toJSON(needs) }}
  - run: ${{ steps.is_success.outputs.result }} # which should evaluate to `true` or `false` and have that status code `1` or `0`
```

which actually looks good is more expressive.


the `notify-build-status` takes the current job context and sends the correct status to slack dependening on the job result. So this would become:

```yaml
  - id: is_success
    uses: elastic/apm-pipeline-library/.github/actions/are-dependent-jobs-success@current
    with: ${{ toJSON(needs) }}
  - run: ${{ steps.is_success.outputs.result }} # which should evaluate to `true` or `false` and have that status code `1` or `0`
  - if: always()
    uses: elastic/apm-pipeline-library/.github/actions/notify-build-status@current
```

reakaleek: https://github.com/elastic/apm-pipeline-library/pull/2126/commits/d208e86c20acaf5687d209714af2c572962c8dbe

.github/actions/are-dependent-jobs-successful/action.yml
```diff
@@ -0,0 +1,21 @@
...
+name: are-dependent-jobs-successful
+description: |
+  Evaluates the combined the status results of the provided needs context.
+inputs:
+  needs:
+    required: true
+    description: needs context as JSON string
+outputs:
+  isSuccess:
+    description: the evaluated result of all provided jobs in the needs context.
```
v1v, line 12: I've just found this output name a bit too verbose, if the action name is `are-dependent-jobs-successful` then the output could be `result`, since success is twice in the action and output name.

What do you think?

On the other hand, I've just though the below

if `are-dependent-jobs-successful` can be changed to `check-dependent-jobs` then it could provide more than one output:
- `success`
- `failure`
- `aborted`

and

```
    - id: test
      run: |
        RESULT=$(echo '${{ inputs.needs }}' | jq -s 'map(.[].result) | all(.=="success")')
        echo "success=${RESULT}" >> $GITHUB_OUTPUT
        RESULT=$(echo '${{ inputs.needs }}' | jq -s 'map(.[].result) | all(.=="failure")')
        echo "failure=${RESULT}" >> $GITHUB_OUTPUT
        RESULT=$(echo '${{ inputs.needs }}' | jq -s 'map(.[].result) | all(.=="abort")')
        echo "aborted=${RESULT}" >> $GITHUB_OUTPUT
      shell: bash
```

Do you think that's something we could use in the future by generalising the action to support different status?

kuisathaverat (approved):

.github/actions/are-dependent-jobs-successful/action.yml
reakaleek: > I've just found this output name a bit too verbose, if the action name is are-dependent-jobs-successful then the output could be result, since success is twice in the action and output name.

my thinking here was that `isSuccess` is more expressive than `result`. WIth the `is` prefix I already know I can expect a boolean. As of `result` could potentially be anything and as a consumer I would need to look deeper into the docs. 

so if we rename it to `check-dependent-jobs` naming something other than `result` makes sense again.

with your suggestion of haven multiple outputs, I'm worried that `all(.=="failure")` is actually an edge case.
I doubt that this will happen in most cases.

so my suggestion would be:

- rename it to `check-dependent-jobs`
- keep the output name `isSuccess`

and if we see usecases for `isFailure` or `isAborted` when can add it in the future.

v1v: > so my suggestion would be:
rename it to check-dependent-jobs
keep the output name isSuccess

I like that approach! Thanks again for iterating on this together!

reakaleek: https://github.com/elastic/apm-pipeline-library/pull/2126/commits/1659d78dc23d38bf3af0b7c4b20c0c056979e656

.github/actions/check-dependent-jobs/action.yml
```diff
@@ -0,0 +1,21 @@
...
+  isSuccess:
+    description: the evaluated result of all provided jobs in the needs context.
+    value: ${{ steps.test.outputs.isSuccess }}
+runs:
+  using: composite
+  steps:
+    - id: test
+      run: |
+        RESULT=$(echo '${{ inputs.needs }}' | jq -s 'map(.[].result) | all(.=="success")')
+        echo "isSuccess=${RESULT}" >> $GITHUB_OUTPUT
```
v1v, line 20: As far as I see `true`/`false` works quite well with the *nix commands as they are recognised commands, while `0`/`1` are not the case.

Do you think converting those values to `true`/`false` could work nicely with the consumers? IIUC, in the example section defined in the README.md uses ` - run: ${{ steps.is_success.outputs.isSuccess }}` so, it will fail in both cases if `0`/`1` while it will work fine with `true`/`false`.

reakaleek: the value of `$RESULT` should be `true` or `false` already. Or am I misunderstanding?

reakaleek: image

Source: https://github.com/elastic/apm-pipeline-library/actions/runs/4362622323/jobs/7627736701

v1v: ahh!! I misread the comment `# should exit with 1 or 0.`, my bad! Then all is good!

v1v (approved):

.github/actions/check-dependent-jobs/README.md
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-latest
+    needs:
+      - job-a
+      - job-b
+    steps:
+      - id: check
+        uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
+        with:
+          needs: ${{ toJSON(needs) }}
+      - run: ${{ steps.is_success.check.isSuccess }} # should exit with 1 or 0.
```
reakaleek, line 47:
```suggestion
      - run: ${{ steps.check.isSuccess }} # should exit with 1 or 0.
```

.github/actions/notify-build-status/README.md
```diff
@@ -69,15 +69,18 @@ jobs:
...
       - release
     runs-on: ubuntu-latest
     steps:
-      - run: test $(echo '${{ toJSON(needs) }}' | jq -s 'map(.[].result) | all(.=="success")') = 'true'
-      - if: always()
-        uses: elastic/apm-pipeline-library/.github/actions/notify-build-status@current
+      - id: is_success
+        uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
+        with: ${{ toJSON(needs) }}
+      - run: ${{ steps.is_success.outputs.isSuccess }}
```
reakaleek, line 82:
```suggestion
      - id: check
        uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
        with: ${{ toJSON(needs) }}
      - run: ${{ steps.check.outputs.isSuccess }}
```

.github/actions/check-dependent-jobs/action.yml
reakaleek: FYI, also renamed the ids: https://github.com/elastic/apm-pipeline-library/pull/2126/files/1659d78dc23d38bf3af0b7c4b20c0c056979e656..8ac2a651d5d842476b515d8b7ead4c264b411485

.github/actions/check-dependent-jobs/README.md
```diff
@@ -40,11 +40,11 @@ jobs:
       - job-a
       - job-b
     steps:
-      - id: is_success
+      - id: check
         uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
         with:
           needs: ${{ toJSON(needs) }}
-      - run: ${{ steps.is_success.outputs.isSuccess }} # should exit with 1 or 0.
+      - run: ${{ steps.check.isSuccess }} # should exit with 1 or 0.
```
reakaleek, line 47:
```suggestion
      - run: ${{ steps.check.outputs.isSuccess }} # should exit with 1 or 0.
```

PR #2128, opened 2023-03-07
action: notify when updatecli failed

v1v:
## What does this PR do?

Notify if the `updatecli` failed with some customisation 

## Why is it important?

Proactive failure notification

### DIFF
--- a/.github/workflows/test-updatecli.yml
+++ b/.github/workflows/test-updatecli.yml
@@ -1,0 +1,20 @@
+---
+name: test-updatecli1
+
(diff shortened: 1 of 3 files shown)


.github/workflows/test-updatecli.yml
```diff
@@ -0,0 +1,20 @@
+---
```
v1v, line 1: This will help us to test these changes once it's merged and before the release happens

reakaleek (approved):

PR #2174, opened 2023-04-04
Publish apm-agent-python docker images in dockerhub

reakaleek:
## What does this PR do?

Publishes apm-agent-python docker images for testing in dockerhub so it can be accessed publicly in GH Actions workflows.

## Why is it important?

Optimizes the test time

## Related Issues
- depends on https://github.com/elastic/apm-agent-python/pull/1794

### DIFF
--- a/.ci/.docker-images.yml
+++ b/.ci/.docker-images.yml
@@ -17,7 +17,7 @@ templates:
     repository: "elastic/apm-agent-python"
     working_directory: "tests"
     build_script: docker build --build-arg PYTHON_IMAGE=${TAG/-/:} -t elasticobservability/apm-agent-python:${TAG} .
-    push_script: docker push elasticobservability/apm-agent-python:${TAG} .
+    push_script: docker push elasticobservability/apm-agent-python:${TAG}
 
   - &apm-agent-nodejs-docker-images
     name: "apm-agent-nodejs"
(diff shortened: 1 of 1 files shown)


v1v (commented):
As far as I see, those python versions are explicitly set in this file, if for instance, python `3.6` reaches EOL or it's removed from  [.matrix_python_full.yml](https://github.com/elastic/apm-agent-python/blob/main/.ci/.matrix_python_full.yml) or if `3.12` is added to that file, then it's required to change this file.
Do you think we can use something more dynamic?

reakaleek:
> As far as I see, those python versions are explicitly set in this file, if for instance, python `3.6` reaches EOL or it's removed from [.matrix_python_full.yml](https://github.com/elastic/apm-agent-python/blob/main/.ci/.matrix_python_full.yml) or if `3.12` is added to that file, then it's required to change this file. Do you think we can use something more dynamic?

Use let me draft something

.ci/.docker-images.yml
```diff
@@ -16,6 +16,8 @@ templates:
     name: "apm-agent-python"
     repository: "elastic/apm-agent-python"
     working_directory: "tests"
+    build_script: docker build --build-arg PYTHON_IMAGE=${TAG/-/:} -t elasticobservability/apm-agent-python:${TAG} .
```
beniwohli, line 19: Just to make sure nobody mistakes these images for actual release artifacts of the agent itself, we could rename them to something like `apm-agent-python-testing`

beniwohli (approved):
Awesome!

v1v (approved):

PR #2176, opened 2023-04-05
remove obsoleted branches from the files with the list of branches from the unified release

v1v:
## What does this PR do?

Remove obsoleted branches from the source of truth.

## Why is it important?

The `artifacts-api` is the one providing the details, but release brnaches that are obsoleted need to be removed, for such I used the below algorithm:

- Given the `branches.json` then get the latest branch (after removing `main`)
- Gets its version and if its patch is different than `0` then
- Remove the minor-1.

For instance, given

```json
{
  "branches": [
    "7.17",
    "8.6",
    "8.7",
    "main"
  ]
}
```

And `8.7` is the latest release branch with the version `8.7.1-SNAPSHOT` 
Since its patch is `1` rather than `0`
Then `8.6` will be removed from the above json file

## Related issues
Closes #ISSUE

### DIFF
--- a/.ci/generate-snapshots.sh
+++ b/.ci/generate-snapshots.sh
@@ -79,6 +79,6 @@ if [ "${searchVersion}" != "${searchLatestBranch}.0"  ] ; then
   ## manipulate minorVersion to get the -1
   newMinorVersion=$(echo "$minorVersion - 1" | bc)
   export removeBranch="${majorVersion}.${newMinorVersion}"
-  jq -r ".branches | map(select(. != env.removeBranch))" branches.json > branches.json.tmp
+  jq -r 'del(.branches[] | select(test(env.removeBranch)))' branches.json > branches.json.tmp
   mv branches.json.tmp branches.json
 fi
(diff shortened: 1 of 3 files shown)


.ci/generate-snapshots.sh
```diff
@@ -0,0 +1,84 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+#
+# It queries the artifacts-api entry point to fetch the next release
+# version for the 8.x line.
```
v1v, line 21:
```suggestion
# It produces the list of active release branches using the artifacts-api and does some
# manipulation to discard those branches that are not active anymore. A bit opinionated. 
```

reakaleek (approved):
LGTM.

But I was wondering if the script would be more readable if we would use python or some other language. 😅

v1v:
> But I was wondering if the script would be more readable if we would use python or some other language. 😅

That's a good point, we probably can use something else instead, so I'll raise an issue so it can be done in a follow up

PR #2194, opened 2023-04-19
ci: add exclude branches on elastic stack snapshot branches 

amannocci:
## What does this PR do?

* Add exclude branches input on elastic stack snapshot branches action.
* Refactor the script in a separate file.

## Why is it important?

* Some use cases require excluding branches from the list of active branches.

### DIFF
--- a/.github/actions/elastic-stack-snapshot-branches/action.yml
+++ b/.github/actions/elastic-stack-snapshot-branches/action.yml
@@ -1,6 +1,12 @@
 ---
 name: 'Elastic Stack Snapshot branches'
 description: 'Fetch the current list of active branches with snapshots in Elastic'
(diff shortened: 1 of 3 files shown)


.github/actions/elastic-stack-snapshot-branches/action.yml
```diff
@@ -1,26 +1,24 @@
...
+    required: false
+    type: string
+    default: ''
 outputs:
   matrix:
     description: "Processed matrix with the required tuples of version and framework"
     value: ${{ steps.generator.outputs.matrix }}
 runs:
   using: "composite"
   steps:
```
v1v, line 16: Good catch!

v1v (approved):

PR #2201, opened 2023-04-28
ci: github action to validate if a PR is approved

amannocci:
## What does this PR do?

* Add a new GitHub action to validate if a PR is approved.
* Implementation testing is achieved using GitHub Actions emulation. 

## Why is it important?

* We needed a way to validate if a PR is approved.
* A PR is approved only if the PR met some criteria.


.github/actions/is-pr-approved/README.md
```diff
@@ -0,0 +1,134 @@
...
+## Features
+* Validate if a PR is approved to run within a workflow.
+* Support default values extracted from context.
+* Support Github PAT token.
+* Support explicit allowed actors.
+
+## Motivation & Context
+
+* We needed a way to check if a PR is approved to run within a workflow.
+* It enable the usage of `pull_request_target` event for forked repositories.
```
v1v, line 21:
```suggestion
* It enables the usage of `pull_request_target` event for forked repositories.
```

.github/actions/is-pr-approved/README.md
```diff
@@ -0,0 +1,134 @@
...
+* The solution must allow a PR to run within the workflow if one of the following conditions is met:
+  * The workflow has been trigger by an allowed actor (explicit list of actors).
+  * The workflow has been trigger by an actor with write/admin permissions on the repository.
+  * The workflow has been trigger by an org actor.
```
v1v, line 25:
```suggestion
  * The workflow has been triggered by an allowed actor (explicit list of actors).
  * The workflow has been triggered by an actor with `write/admin` permissions on the repository.
  * The workflow has been triggered by an org actor.
```

.github/actions/is-pr-approved/src/main.ts
```diff
@@ -0,0 +1,161 @@
...
+import {GitHub} from '@actions/github/lib/utils'
+import {actionInputs, transformIfSet} from 'github-actions-utils'
+
+interface GitHubAPIError {
+  status: number
+  request: any // eslint-disable-line @typescript-eslint/no-explicit-any
+  response: any // eslint-disable-line @typescript-eslint/no-explicit-any
+}
+
+export const DEFAULT_ALLOWED_ACTORS: readonly string[] = [
```
v1v, line 12: NIT:
What do you think using an input parameter with this list of BOT actors rather than being hardcoded? This can help to be consumed this action by others and configure those actors if needed.

.github/actions/is-pr-approved/src/main.ts
```diff
@@ -0,0 +1,161 @@
...
+export async function run(): Promise<void> {
+  // Use dynamic import to let `github-action-ts-run-api` library inject
+  // Github Context before using it.
+  // Static import will not work and result in empty Github Context
+  const github = await import('@actions/github')
+
+  try {
+    // Validate event source
+    if (!['pull_request', 'pull_request_target'].includes(github.context.eventName)) {
+      return core.setFailed('This action only support `pull_request` and `pull_request_target` events')
```
v1v, line 28: Given the below GitHub action:

```yaml

on:
  workflow_dispatch:
  pull_request:
  push:
    branches: [ 'main']
...
```

Will it fail when running a manual build using the GitHub UI or when merging into `main`?

v1v (commented):
Maybe dependabot could be configured for this particular action/ts?

.github/actions/is-pr-approved/src/main.ts
amannocci: Currently, both will fail due to the specific PR review logic.

.github/actions/is-pr-approved/src/main.ts
amannocci: It's possible to override the list of allowed actors using the `allowed-actors` input.
However, Hardcoding default allowed actors in the code rather than in the `actions.yml` for test purposes is easier.

.github/actions/is-pr-approved/src/main.ts
amannocci: I added support for those events. Do we need to rename the action to something like `is-workflow-approved` to reflect the support of those events?

.github/actions/is-pr-approved/src/main.ts
```diff
@@ -0,0 +1,167 @@
...
+  const github = await import('@actions/github')
+
+  try {
+    // Validate event source
+    const eventName = github.context.eventName
+    if (['push', 'workflow_dispatch'].includes(eventName)) {
+      return core.info(`Skipping validation due to priviledge event (${eventName})`)
+    } else if (!['pull_request', 'pull_request_target'].includes(eventName)) {
+      return core.setFailed(UNSUPPORTED_EVENT_ERROR)
+    }
```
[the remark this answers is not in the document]
amannocci: This action should only allow actors that have enough privilege to run specific jobs or actions.
There are some cases where implicitly we know that the actor has enough privilege (`push` or workflow_dispatch event).
However, there are some cases where we can't implicitly know if the actor has enough privilege.
For example, this action can be run with `pull_request_review_comment` and bypass checks.
This could lead to unexpected behavior in some cases.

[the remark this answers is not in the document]
amannocci: Perhaps, the scope of this action should be redefined.
The scope is either a generic workflow or a workflow specific for pull_request/pull_request_target.
The goal of this action is to authorise an actor to perform some actions within a workflow based on the GitHub context.
From a user perspective, we could assume that this action can be used in any workflow and triggered by any event.
In that case, we can support some events where we can safely assume that the actor is authorised and otherwise deny by failing.
In addition, we can add an input to skip the validation for an explicit list of events.
WDYT about the above?

P.S: is-actor-authorise or similar seems a better name in this context.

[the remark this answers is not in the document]
amannocci: AFAIK, there is only the `issue_comment` event that can be interesting to support.
In that case, the action could be used to check if the actor is authorized to trigger the workflow.
Otherwise, we could assume that the workflow was triggered by an authorized actor and skip the validation.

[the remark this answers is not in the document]
v1v: > there is only the issue_comment event that can be interesting to support.

That particular event won't need to be supported atm (see the internal issue linked to https://github.com/elastic/apm-agent-nodejs/pull/3258 that explains a bit the reason)

.github/actions/is-pr-approved/src/main.ts
amannocci: The default list of allowed actors is present in the `How to use it` section.

v1v (approved):

PR #2205, opened 2023-05-02
snapshots: remove majors if there are 3 versions

v1v:
## What does this PR do?

Remove branches if for the same major version there are 3 elements

## Why is it important?

Artifacts API does not care about active release branches, hence it's required to delete the one is not supported.

https://github.com/elastic/apm-pipeline-library/pull/2176 was one case to care but there is also another one when an old branch is generated then it will be shown in the `artifacts-api`.

### DIFF
--- a/.ci/generate-snapshots.sh
+++ b/.ci/generate-snapshots.sh
@@ -28,17 +28,21 @@ cd snapshots
 URL="https://artifacts-api.elastic.co/v1"
 NO_KPI_URL_PARAM="x-elastic-no-kpi=true"
 
+
+echo ">> Query versions in ${URL}"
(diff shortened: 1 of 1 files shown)


.ci/generate-snapshots.sh
```diff
@@ -68,6 +73,17 @@ for branch in $(jq -r '.branches | .[]' branches.json); do
   fi
 done
 
+echo ">> No more than 2 versions for the same minor is needed"
+if [ "$(jq -r '.branches | map(select(. | startswith("8."))) | length' branches.json)" == "3" ] ; then
```
v1v, line 77: This is the new logic, as today:

```json
{
  "branches": [
    "7.17",
    "8.6",
    "8.7",
    "8.8",
    "main"
  ]
}

```


but `8.6` is not an active release branch, hence it should not be in the list of active branches

reakaleek (approved):

PR #2208, opened 2023-05-03
updatecli: add what CI workflow generated the PR

v1v:
## What does this PR do?

Add what workflow generated the PR

## Why is it important?

I see some duplicated PRs, so this might help in addition this can help contributors so they know how things are generated.


reakaleek (approved):
LGTM

Looks like we could maybe create an action `setup-common-env` that exports predefined environment variables that we commonly use.

(just an idea)

PR #2216, opened 2023-05-15
build: correct pre-commit lint

amannocci:
## What does this PR do?

* Correct pre-commit lint configuration after #2201

## Why is it important?

* Avoid pre-commit false positive errors.


v1v (approved):
Thanks! 🙏

PR #2225, opened 2023-05-23
test-action: disable emoji reaction for now

v1v:
Unrequired!

### DIFF
--- a/.github/workflows/test-comment-reaction.yml
+++ b/.github/workflows/test-comment-reaction.yml
@@ -4,7 +4,9 @@ name: test-comment-reaction
 on:
   issue_comment:
     types:
-      - created
+      # disable for the time being...
+      #- created
+      - deleted
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-comment-reaction.yml
```diff
@@ -4,7 +4,9 @@ name: test-comment-reaction
 on:
   issue_comment:
     types:
-      - created
+      # disable for the time being...
+      #- created
+      - deleted
```
reakaleek, line 9: Curious: if it's meant to be disabled, why add the `deleted` event?

reakaleek (approved):

.github/workflows/test-comment-reaction.yml
reakaleek: We could also do: https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow

PR #2238, opened 2023-06-09
refactor: improve github action slack message

amannocci:

* Refactor Python code using typescript implementation.

* Use standard GitHub action libraries.
* Support properly multiline without explicit URL encoding.
* Add integration tests and e2e tests.

## How has this been tested?

* There are integration tests runnable using `npm test`.
* There are e2e tests in a specific GitHub Workflow.

## How has this been documented?

* README.md updated


.github/actions/slack-message/src/main.ts
```diff
@@ -0,0 +1,80 @@
...
+  message = decodeURIComponent(message)
+
+  // Ref: https://api.slack.com/reference/surfaces/formatting#escaping
+  message = message.replace('&', '&amp;')
+  message = message.replace('<', '&lt;')
+  message = message.replace('>', '&gt;')
+
+  // Support multilines
+  return message.split(/\r?\n/).join('\n')
+}
```
kuisathaverat, line 80: Why not URL encode the string? Sanitize custom functions usually are incomplete, error-prone, and end on security issues.

amannocci: Is there a limitation on the GitHub Action side regarding multiline input?
The function sanitize the message according to the Slack documentation.
In addition, the output var formatting is handled within the `github/core` library.

PR #2256, opened 2023-06-19
Make inputs.message optional

reakaleek:
## What does this PR do?

FIxes the notify-build-status action.

After the refactor the action failed when  `inputs.message` is not provided.

However, the message should be optional.

Example: https://github.com/elastic/opbeans-java/actions/runs/5308382176/jobs/9607800679

## Why is it important?

All notifications without a message are failing.

## Related issues
Closes #ISSUE


.github/actions/notify-build-status/src/main.ts
```diff
@@ -13,7 +13,7 @@ export async function run(): Promise<void> {
     const githubSha = checkEnvVar('GITHUB_SHA')
     const githubWorkflow = checkEnvVar('GITHUB_WORKFLOW')
     const jobStatus = checkEnvVar('JOB_STATUS')
-    const message = checkEnvVar('MESSAGE')
+    const message = process.env.MESSAGE || undefined
```
amannocci, line 16:
```suggestion
    const message = process.env.MESSAGE || ''
```
Can we use an empty string instead of an undefined one to force the var type?

amannocci:
@reakaleek, can you correct the lint? The CI isn't running since we switch from `pull_request` to `pull_request_target`.

reakaleek:
> @reakaleek, can you correct the lint? The CI isn't running since we switch from `pull_request` to `pull_request_target`.

I splitted the workflow into 2 workflows with different triggers. Hence, the wrapper test is executed normally now.

.github/workflows/test-notify-build-status-integration.yml
```diff
@@ -0,0 +1,59 @@
+---
+name: test-notify-build-status-integration
+
+on:
+  pull_request_target:
+    branches:
+      - main
+    paths:
```
amannocci, line 8: Does it make sense to add the workflow itself?

.github/workflows/test-notify-build-status-integration.yml
```diff
@@ -0,0 +1,59 @@
...
+      - '.github/actions/notify-build-status/**'
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
```
amannocci, line 14: Does it make sense to add the workflow itself?

.github/workflows/test-notify-build-status.yml
```diff
@@ -2,16 +2,19 @@
...
 on:
-  workflow_dispatch:
   pull_request:
     paths:
       - '.github/actions/notify-build-status/**'
+  workflow_dispatch:
   push:
     branches:
       - main
     paths:
```
amannocci, line 12: Does it make sense to add the workflow itself?

.github/workflows/test-notify-build-status.yml
```diff
@@ -2,16 +2,19 @@
 name: test-notify-build-status
 
 on:
-  workflow_dispatch:
   pull_request:
     paths:
```
amannocci, line 6: Does it make sense to add the workflow itself?

amannocci (approved):
LGTM
I let some comments.

.github/workflows/test-notify-build-status.yml
reakaleek: https://github.com/elastic/apm-pipeline-library/pull/2256/commits/bd5129ee259a4d3d9bc95c14acc78775e2674b1a

.github/workflows/test-notify-build-status.yml
reakaleek: https://github.com/elastic/apm-pipeline-library/pull/2256/commits/bd5129ee259a4d3d9bc95c14acc78775e2674b1a

.github/workflows/test-notify-build-status-integration.yml
reakaleek: https://github.com/elastic/apm-pipeline-library/pull/2256/commits/bd5129ee259a4d3d9bc95c14acc78775e2674b1a

amannocci (approved):

PR #2313, opened 2023-07-28
ci: update dependencies and remove sourcemap

amannocci:
* Upgrade dependencies
* Remove sourcemap since we aren't using it

* Maintenance


reakaleek (approved):

.github/actions/snapshoty/.node-version
reakaleek, line 1: can this be an `.nvmrc` file?

PR #2332, opened 2023-08-15
Update nodeJSDefaultVersion.groovy to 18.17.1

andrewvc:
## What does this PR do?

Upgrades the node version to the latest LTS version, this was breaking synthetics tests that depend on node 14 features.
I'm not sure if other Jenkins stuff would have an issue with this version.

## Why is it important?

Unblocks development of heartbeat

## Related issues

https://github.com/elastic/beats/pull/36147/

### DIFF
--- a/vars/nodeJSDefaultVersion.groovy
+++ b/vars/nodeJSDefaultVersion.groovy
@@ -30,5 +30,5 @@ def call(Map args = [:]) {
 }
 
 def defaultVersion(){
-  return '14.17.5'
+  return '18.17.1'
 }


andrewvc:
@cachedout anything we need to do to make sure the next version of this in build kite just pulls the latest node LTS rather than a fixed version?

v1v:
> make sure the next version of this in build kite just pulls the latest node LTS rather than a fixed version?

I'm not aware of how it's implemented in Buildkite. Can you explain your use case?

On the other hand, this particular fix won't be needed as long as there is a `.nvmrc` file in the root of the project, then this particular step will [read it](https://github.com/elastic/apm-pipeline-library/blob/7dada357c95aaee674ac661cd44969bb20b32073/vars/nodeJSDefaultVersion.groovy#L25)

v1v (approved):

v1v:
I've just released a new version of this library, https://github.com/elastic/apm-pipeline-library/releases/tag/v1.1.404. Son consumers can now use this change - there is nothing else to be done, since consumers of this library use the `current` tag and that particular tag has been updated to `v1.1.404`.

PR #2366, opened 2023-09-19
action: refactor serverless params

v1v:
## What does this PR do?

Use script for preparing the oblt-cli variables

## Why is it important?

Platform agnostic

## Related issues
Closes #ISSUE

## Test

image

### DIFF
--- a/.github/actions/oblt-cli-create-serverless/action.yml
+++ b/.github/actions/oblt-cli-create-serverless/action.yml
@@ -37,71 +37,42 @@ inputs:
 runs:
   using: "composite"
   steps:
-
     - name: Configure oblt-cli flags
-      run: |
-        {
-          [ -n '${{ inputs.cluster-name-prefix }}' ] \
-            && echo -n '--cluster-name-prefix=${{ inputs.cluster-name-prefix }} '
-          [ -n '${{ inputs.cluster-name-suffix }}' ] \
-            && echo -n '--cluster-name-suffix=${{ inputs.cluster-name-suffix }} '
-          [ '${{ inputs.dry-run }}' == 'true' ] \
-            && echo -n '--dry-run '
-        } > .flags
-        echo "COMMAND=$(cat .flags)" >> $GITHUB_ENV
-        rm .flags
-      shell: bash
+      uses: actions/github-script@v6
+      with:
+        script: |
+          const { cluster-name-prefix, cluster-name-suffix, dry-run, gitops, target, project-type  } = process.env
+          if (cluster-name-prefix != '') {
+            core.exportVariable('CLUSTER_NAME_PREFIX', `--cluster-name-prefix==${cluster-name-prefix}`)
+          }
+          if (cluster-name-prefix != '') {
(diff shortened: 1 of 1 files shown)


.github/actions/oblt-cli-create-serverless/action.yml
```diff
@@ -37,71 +37,50 @@ inputs:
...
+                parameters.GitHubIssue = context.issue.number
+              }
+            } else {
+              parameters.GitHubPullRequest = context.issue.number
+            }
+          }
+          console.log(`Create PARAMETERS env variable`)
+          core.exportVariable('PARAMETERS', `--parameters='${(JSON.stringify(parameters))}'`)
+      env:
+        cluster_name_prefix: ${{ inputs.cluster-name-prefix }}
```
v1v, line 74: I could not use `process.env` since those variables contain some `hyphens` and somehow I faced some errors. So I decided to use the env approach to solve my initial problem and also to protect from any input injections, as explained in https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

.github/actions/oblt-cli-create-serverless/action.yml
```diff
@@ -37,71 +37,50 @@ inputs:
...
-      shell: bash
-
-    - name: Configure oblt-cli flags for gitops - pull-request
-      if: github.event.issue.pull_request && contains(inputs.gitops, 'true')
-      run: |
-        {
-          echo -n '{'
-          echo -n '"GitOps":"true",'
-          echo -n '"GitHubRepository":"${{ github.repository }}",'
-          echo -n '"GitHubCommit":"${{ github.event.comment.id }}",'
```
v1v, line 76: This is not supported, I could not find where we use it so far. And I could not find how to gather the sha commit. Somehow I feel we don't use it and we don't need it anyway

reakaleek (approved):
LGTM,

in general, I was wondering, if we could put those github-scripts into a JS file, so we can benefit from syntax highlighting and code completion, thus increasing the maintainability?

v1v:
> if we could put those github-scripts into a JS file
I could not find how to use `${{ github.action_path }}` in https://github.com/actions/github-script . But it seems I can use `GITHUB_ACTION_PATH`

Let me try again

reakaleek:
> > if we could put those github-scripts into a JS file
> 
> I could not find how to use `${{ github.action_path }}` in https://github.com/actions/github-script . But it seems I can use `GITHUB_ACTION_PATH`
> 
> Let me try again

I found this: https://github.com/actions/github-script#run-a-separate-file

reakaleek:
> > > if we could put those github-scripts into a JS file
> > 
> > 
> > I could not find how to use `${{ github.action_path }}` in https://github.com/actions/github-script . But it seems I can use `GITHUB_ACTION_PATH`
> > Let me try again
> 
> I found this: https://github.com/actions/github-script#run-a-separate-file

Actually, this seems to make it more complicated.

> Additionally, you'll want to use the [checkout action](https://github.com/actions/checkout) to make sure your script file is available.

PR #2372, opened 2023-09-21
Add ability to cancel BK build when workflow is cancelled

reakaleek:
## What does this PR do?

Add ability to cancel BK build when workflow is cancelled

## Why is it important?

When we have a long running BK build and `waitFor == true` and we want to cancel the workflow.

## Related issues
Closes #ISSUE

### DIFF
--- a/.github/actions/buildkite/action.yml
+++ b/.github/actions/buildkite/action.yml
@@ -112,7 +112,7 @@ runs:
           echo "BUILDKITE_API_ACCESS_TOKEN=" >> $GITHUB_ENV
 
       - name: Cancel Buildkite pipeline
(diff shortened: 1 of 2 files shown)


.github/actions/buildkite/action.yml
```diff
@@ -110,3 +110,13 @@ runs:
         shell: bash
         run: |
           echo "BUILDKITE_API_ACCESS_TOKEN=" >> $GITHUB_ENV
+
+      - name: Cancel Buildkite pipeline
+        if: cancelled() inputs.waitFor && steps.trigger-buildkite.outputs.build_number
```
v1v, line 115:
```suggestion
        if: cancelled() && inputs.waitFor && steps.trigger-buildkite.outputs.build_number
```

v1v (approved):

PR #2377, opened 2023-09-26
actions: manage github branches

v1v:
## What does this PR do?

GitHub action to handle the locking/unlocking Branch protection given the PAT

## Why is it important?

To help with securing the branch protection and supporting freezing the branch where the releases are happening

### DIFF
--- a/.github/actions/manage-branch/action.yml
+++ b/.github/actions/manage-branch/action.yml
@@ -1,0 +1,25 @@
+name: 'Manage GitHub branches'
+description: 'Manage the given GitHub branches'
+inputs:
+  branch:
+    description: 'The GitHub branch'
+    required: true
+  lock:
+    description: 'Whether to lock the branch or no'
+    required: true
+  token:
+    description: 'The GitHub access token.'
+    required: true
+  owner:
+    description: 'The GitHub owner'
+    default: ${{ github.repository }}
+  repo:
+    description: 'The GitHub repo'
+    default: ${{ github.repository }}
+runs:
+  using: "composite"
+  steps:
+    - run: ${{ github.action_path }}/lock.py ${{ inputs.owner }} ${{ inputs.repo }} ${{ inputs.branch }} ${{ inputs.lock }}
+      env:
+        GITHUB_TOKEN: ${{ inputs.token }}
+      shell: shell
--- a/.github/actions/manage-branch/lock.py
+++ b/.github/actions/manage-branch/lock.py
@@ -1,0 +1,77 @@
+import os
+import requests
+import sys
+
+def post(query):
+  # Get environment variables
+  GITHUB_TOKEN = os.getenv('GITHUB_TOKEN')
+
+  headers = {
+    "Authorization": "Bearer " + GITHUB_TOKEN,
+    "Content-type": "application/json"
+  }
+
+  response = requests.post(
+      url='https://api.github.com/graphql',
+      headers=headers,
+      json={"query": query},
+  )
+
+  return response
+
+def branch_protection_query(owner, repo):
+  return """query {
+    repository(name: "%s", owner: "%s") {
+      branchProtectionRules(first: 100) {
+        nodes { id, pattern }
+      }
+    }
+  }""" % (repo, owner)
+
+def set_lock(protection_id, enabled):
+  return """mutation {
+    updateBranchProtectionRule(input: {
+      branchProtectionRuleId: "%s",
+      lockBranch: %s
+    }){
+      clientMutationId
+    }
+  }""" % (protection_id, enabled)
+
+def process_error(response):
+  if "errors" in response:
+    print('ERROR: ' + response['errors'][0]['message'])
+    sys.exit(1)
+
+def fetch_branch_protections(owner, repo):
+  print('INFO: fetch the branch protections for "%s/%s"' % (owner, repo))
+  response = post(branch_protection_query(owner, repo))
+  process_error(response.json())
+  return response.json()['data']['repository']['branchProtectionRules']['nodes']
+
+def update_lock(protection_id, lock):
+  print('INFO: setting lock("%s")'% (enabled))
+  query = set_lock(protection_id, str(enabled).lower())
+  response = post(query)
+  process_error(response.json())
+  print('INFO: all done!')
+
+# If no arguments then say it!
+if len(sys.argv) < 5:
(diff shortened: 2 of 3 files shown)


amannocci:
Superseded by https://github.com/elastic/apm-pipeline-library/pull/2432

amannocci:
## Vulnerability
The current implementation is vulnerable to craft injection.

For example, it's easy to remove a branch protection:
```yaml
      - uses: elastic/apm-pipeline-library/.github/actions/manage-branch@feat/python-branch-lock
        with:
          lock: '"false, pattern: \"foobar\""'
          branch: 'main'
          token: ${{ secrets.PAT }}
```
The above execution will replace the branch protection on the main branch and replace it with branch protection on the foobar branch.

It's also possible to craft a malicious request that contains multiple mutations.
In that case, we can virtually interact with anything that the GitHub Personal Token is allowed to.

amannocci (changes requested):
We need to take care of the vulnerability

v1v:
> Need to take care of the vulnerability

Thanks for this! This is a draft PR to illustrate how to use the existing graphql-api. Contributions are welcome 👼 

Even though I understand the theory about the attack-vector, I cannot think of any genuine reasons this will even happen by any malicious actor. As far as I'm concerned this type of action will be used by consumers with some hardcoded values rather than dynamic values. To change the call to this action will require to fulfil the branch protection policies through a Pull Request, hence the possibility is very low but the probability based on this particular scenario is certainly zero

For the shake of mind, we can put as many validations as needed if that can make this more secured, though, let's not forget  about the fundamental of this action about supporting releases with some branch protections.

.github/actions/manage-branch/lock.py
```diff
@@ -0,0 +1,94 @@
...
+  return """query {
+    repository(name: "%s", owner: "%s") {
+      branchProtectionRules(first: 100) {
+        nodes { id, pattern }
+      }
+    }
+  }""" % (repo, owner)
+
+def set_lock(protection_id, enabled):
+  return """mutation {
```
v1v, line 32:
```suggestion
  if enabled in ("True", "true", True, "False", "false", False):
    print('::error::<enabled> is not a boolean type')
    sys.exit(1)
  return """mutation {
```

kuisathaverat:
> ## Vulnerability
> The current implementation is vulnerable to craft injection.
> 
> For example, it's easy to remove a branch protection:
> 
> ```yaml
>       - uses: elastic/apm-pipeline-library/.github/actions/manage-branch@feat/python-branch-lock
>         with:
>           lock: '"false, pattern: \"foobar\""'
>           branch: 'main'
>           token: ${{ secrets.PAT }}
> ```
> 
> The above execution will replace the branch protection on the main branch and replace it with branch protection on the foobar branch.
> 
> It's also possible to craft a malicious request that contains multiple mutations. In that case, we can virtually interact with anything that the GitHub Personal Token is allowed to.

The following step deleted the GitHub Action protection, once you have a `PAT` you can do whatever you want, I do not see your point.

```bash
      - run: |
          gh api -X DELETE repos/owner/repo/branches/main/protection
        env:
          GITHUB_TOKEN: ${{ secrets.PAT }}

```

amannocci:
> > ## Vulnerability
> > The current implementation is vulnerable to craft injection.
> > For example, it's easy to remove a branch protection:
> > ```yaml
> >       - uses: elastic/apm-pipeline-library/.github/actions/manage-branch@feat/python-branch-lock
> >         with:
> >           lock: '"false, pattern: \"foobar\""'
> >           branch: 'main'
> >           token: ${{ secrets.PAT }}
> > ```

> > The above execution will replace the branch protection on the main branch and replace it with branch protection on the foobar branch.
> > It's also possible to craft a malicious request that contains multiple mutations. In that case, we can virtually interact with anything that the GitHub Personal Token is allowed to.
> 
> The following step deleted the GitHub Action protection, once you have a `PAT` you can do whatever you want, I do not see your point.
> 
> ```shell
>       - run: |
>           gh api -X DELETE repos/owner/repo/branches/main/protection
>         env:
>           GITHUB_TOKEN: ${{ secrets.PAT }}
> ```

Yeah sure but in the above case, we don't introduce any new step within the workflow.
In my opinion, the input of any GitHub Action should be considered unsafe.
We can imagine a user (Open Source Project) that passes dynamic parameters issued from inputs or any GitHub context.

kuisathaverat:
>In my opinion, the input of any GitHub Action should be considered unsafe.

Agree, that any user input must be considered unsafe and not processed directly, check the type of the value is correct and sanitize always.

.github/actions/manage-branch/action.yml
```diff
@@ -0,0 +1,28 @@
...
+  repo:
+    description: 'The GitHub repo'
+    default: ${{ github.repository }}
+  fail-on-warning:
+    description: 'Whether to fail if any warning'
+    default: false
+runs:
+  using: "composite"
+  steps:
+    - run: python ${{ github.action_path }}/lock.py ${{ inputs.owner }} ${{ inputs.repo }} ${{ inputs.branch }} ${{ inputs.lock }} ${{ inputs.fail-on-warning }}
```
kuisathaverat, line 25: it is better to use environment variables and put them between double quotes

v1v: yep, bear in mind, this PR is in draft and was not intended to be even in review, it was only created to illustrate that we could use this approach for freezing/unfreezing branches on demand.

cachedout:
@amannocci asked me to weigh in on this. I've had a look at both PRs and my reaction is that I'm not a very big fan of having deployment workflows modifying branch protection rules. Even with well-tested code, this feels like unnecessary risk and complexity. 

I don't know a lot about the background here but I suspect from the context that this is about a long-running release process in which case it is necessary to ensure that the branch is not modifying during its execution. Assuming that is correct, I would propose an approach that could be much easier.

What if we "lock" the branch using GitHub's own [deployment locking mechanism](https://github.com/github/lock) and combine this with a merge queue that checks this lock and halts its execution if a lock is found. Is there a reason that wouldn't work?

cachedout:
@v1v has kindly pointed me to his comments [here](https://github.com/elastic/observability-robots/issues/1864#issuecomment-1803876914) which appear to be heading in the same direction as my own. My preference would be that we explore that as an option before returning to a discussion about the viability of either of these two PRs.

PR #2381, opened 2023-10-05
fix: create the arm64 Kibana Docker image for serverless

kuisathaverat:
## What does this PR do?

create the arm64 Kibana Docker image for serverless

## Why is it important?

The current implementation only generates amd64 Docker images that are no longer valid for serverless projects in QA. We split the command to better maintainability due to Kibana build flags can impact each other. We need to review if it is possible to skip something else in the serverless build because it takes longer than before.

### DIFF
--- a/.github/actions/kibana-docker-image/build-and-push.sh
+++ b/.github/actions/kibana-docker-image/build-and-push.sh
@@ -48,7 +48,6 @@ time node scripts/build \
       --docker-namespace="${DOCKER_NAMESPACE}" \
       --docker-tag="${DOCKER_TAG}" \
       --docker-push \
-      --skip-archives \
       --skip-initialize \
       --skip-docker-contexts \
       --skip-docker-ubi \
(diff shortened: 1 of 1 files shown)


v1v (dismissed):

.github/actions/kibana-docker-image/build-and-push.sh
```diff
@@ -32,21 +32,32 @@ echo "::endgroup::"
...
+        --skip-initialize \
+        --skip-docker-contexts \
+        --skip-docker-ubi \
+        --skip-docker-ubuntu \
+        --skip-generic-folders \
+        --skip-platform-folders \
+        --skip-docker-serverless
 else
-  skip_docker_flag="--skip-docker-cloud"
+  # enable Docker multiarch support
```
kuisathaverat, line 49:
```suggestion
  # enable Docker multiarch support
```

v1v (approved):

PR #2429, opened 2023-11-08
chore: Loosen up opentelemetry versions

jaysonsantos:
This make sures the downstream can consume this library without pinning otel's version to 1.15.0.
Otel's version is fairly stable so, everything >=1 and <2 should be safe.


kuisathaverat:
this is automated with Dependabot, we use only fixed versions. It requires some changes to make tests pass, I have to find time to do it.

PR #2432, opened 2023-11-08
feat: implement github action to lock/unlock branch

amannocci:

* Implement a new GitHub Action to lock or unlock a branch using existing branch protections.
* Add e2e tests for the GitHub Action.

* We need a way to lock a base branch during release process in some repositories.


v1v (commented):
What's the advantage for using this versus https://github.com/elastic/apm-pipeline-library/pull/2377?

amannocci:
> What's the advantage for using this versus #2377?

### Testability
* There are well-supported libraries to test the GitHub Action in a standard way.

### Maintainability
* We rely on libraries that are provided and maintained by GitHub.

### Security
* Since we rely on GitHub libraries, we have security features for free like commands, input escapes etc..
* There is also a GraphQL query validation to avoid crafting bad request.

v1v:
Thanks for the explanation. 

I see the scope of this action quite reduced in terms of functionality (a graphql query to gather the branch protections and apply the lock/unlock for the given branch), for that reason, I'd like to mention a few more things:

The other PR was in principle made as simple a possible, platform agnostic using native libraries rather than any external dependencies but requests (which it’s available in the CI runners by default) to rely on the GraphQL API with raw http requests. As you spotted, there is a lack of UTs that’s certainly something to be improved - it was not implemented intentionally.

There are different reasons that made me stick to the other PR, such as the number of files, the number of new lines `63,000` ❗, the required dev tools, the dependencies but the system ones, the package idiosyncrasy (to put the files under the list folder using acc) or the need to bump those dependencies often.

While I understand the benefits for using typescript for certain cases, in this particular case, I'd like to know your point of view about contributing to the other PR with some UTs so we could use it instead of this one?

Thanks

amannocci:
## Overview

AFAIS, we have two different approaches.
We can stick with the simplicity of the first PR.
Or we could use the second one for the whole eco-system.

## Python Implementation

### Pros
* Concise cross-platform implementation.
* Relies on already installed software listed as supported (Python & Bash).

### Cons
* Relies on a dependency (requests) not listed as supported (Not explicit).
* Hard to test:
  * It's possible to add UT using pytest, which implies setting up a whole environment using pyproject.toml config file. 
  * It's possible to test the action within a workflow but assertions are hard to describe. 

## Typescript Implementation

### Pros
* Follow standard custom GitHub actions based on GitHub documentation. 
* More secured by nature (GitHub support, validation).
* Easier to test e2e with specific libraries designed to emulate a GitHub Runner in various states.
* Standalone dist using ncc which doesn't need any particular setup.
* Rely on a well-defined node version (granted to be the specified major version).

### Cons
* The GitHub Action is considered as a production-ready software with a full workflow.
* Need to be packaged at build time and pushed within the repository.

## Point of view
* In my opinion, we should treat a GitHub Action as any software and support it with a minimal test suite.
* This implies in both cases to setup an environment (dependencies).
* Since the typescript approach is officially documented and supported by GitHub, I tend to believe that the typescript approach is a better option in the long run.
* However, we could still go with the simplicity (aka Python) and break-fix in the future.
* On the other hand, this GitHub Action is meant to be used in the release process and could lead to disruption if unstable.

v1v:
Thanks for the overview. 

I understand there is an amalgam of implementations but I'd rather step back and analyse the pros/cons altogether so we can be sure we are all aligned.  

To sum up, as long as we don't have an agreement about the foundation for writing and maintaining GitHub actions, let's keep it simple with a well defined contract - certainly this approach has been working well for now. Then we can revisit those implementations in the future once we have agreed so. Probably with this agreement we will come up with some guidelines and procedures that are explicit about this and will be materialised in some documentation within this repository.

.github/actions/branch-lock/src/main.ts
```diff
@@ -0,0 +1,104 @@
+import * as core from '@actions/core'
+import {actionInputs} from 'github-actions-utils'
+import {GraphQLClient, gql} from 'graphql-request'
```
kuisathaverat, line 3: Oktokit already have a library for GrapQL https://github.com/octokit/octokit.js#graphql-api-queries

.github/actions/branch-lock/src/main.ts
```diff
@@ -0,0 +1,104 @@
...
+    // Validate inputs
+    const lock = actionInputs.getBool('lock', true)
+    const branch = actionInputs.getString('branch', true)
+    const token = actionInputs.getString('token', true, true)
+    const owner = actionInputs.getString('owner', true)
+    const repo = actionInputs.getString('repo', true)
+
+    // Create github service
+    const githubService = new GitHubService(token)
+    const branchId = await githubService.searchBranchProtection(owner, repo, branch)
```
kuisathaverat, line 19: This call already exists in oktokit https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/main/docs/repos/getBranchProtection.md

.github/actions/branch-lock/src/main.ts
```diff
@@ -0,0 +1,104 @@
...
+    await githubService.setLock(branchId, lock)
```
kuisathaverat, line 20: dito https://github.com/octokit/plugin-rest-endpoint-methods.js/blob/main/docs/repos/updateBranchProtection.md

kuisathaverat:
Since all the services implemented in this action are implemented already in the oktokit, I do not see the need of reimplement them from scratch, we can use the two methods provided by oktokit to make all the work.

.github/actions/branch-lock/src/main.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+class BranchNotFound extends Error {
+  /**
+   * Init branch not found error.
+   * @param branchName branch name.
+   */
+  constructor(branchName: string) {
+    super(`Failed to find the branch: ${branchName}`)
+  }
+}
```
kuisathaverat, line 100: I did not test it but this core replace your implementation

```suggestion
    var branchProtection = octokit.rest.repos.getBranchProtection({
      owner,
      repo,
      branch,
    });
    branchProtection.lockBranch=lock
    octokit.rest.repos.updateBranchProtection({
        owner,
        repo,
        branch,
        ...branchProtection
      })
    core.info('Succeed to update branch lock.')
  } catch (err: unknown) {
    if (err instanceof Error) {
      core.setFailed(err)
    } else {
      core.setFailed('Unhandled error occured')
    }
  }
}
```

amannocci: Unfortunately, the branch protection response from octokit doesn't match the request body in the updateBranchProtection method.
Assuming we install `octokit/rest` instead of `octokit/core`.

```typescript
    // Create github service
    const githubService = new Octokit({auth: token})
    const response = await githubService.rest.repos.getBranchProtection({
      owner,
      repo,
      branch,
    })
    const branchProtection = response.data
    branchProtection.lock_branch = {enabled: lock}
    await githubService.rest.repos.updateBranchProtection({
      owner,
      repo,
      branch,
      ...branchProtection
    })
```

The above implementation will lead to a compile error at build time.
```txt
src/main.ts:26:59 - error TS2345: Argument of type '{ url?: string | undefined; enabled?: boolean | undefined; required_status_checks?: { url?: string | undefined; enforcement_level?: string | undefined; contexts: string[]; checks: { context: string; app_id: number | null; }[]; contexts_url?: string | undefined; strict?: boolean | undefined; } | undefined; ... 15 mor...' is not assignable to parameter of type 'RequestParameters & Omit<ToOctokitParameters<{ parameters: { path: { owner: string; repo: string; branch: string; }; }; requestBody: { content: { "application/json": { required_status_checks: { ...; } | null; ... 9 more ...; allow_fork_syncing?: boolean | undefined; }; }; }; responses: { ...; }; }>, "baseUrl" | ... ...'.
  Type '{ url?: string | undefined; enabled?: boolean | undefined; required_status_checks?: { url?: string | undefined; enforcement_level?: string | undefined; contexts: string[]; checks: { context: string; app_id: number | null; }[]; contexts_url?: string | undefined; strict?: boolean | undefined; } | undefined; ... 15 mor...' is not assignable to type 'Omit<ToOctokitParameters<{ parameters: { path: { owner: string; repo: string; branch: string; }; }; requestBody: { content: { "application/json": { required_status_checks: { strict: boolean; contexts: string[]; checks?: { context: string; app_id?: number | undefined; }[] | undefined; } | null; ... 9 more ...; allow_...'.
    Types of property 'required_linear_history' are incompatible.
      Type '{ enabled?: boolean | undefined; } | undefined' is not assignable to type 'boolean | undefined'.
        Type '{ enabled?: boolean | undefined; }' is not assignable to type 'boolean | undefined'.
```

We need to pass most of the parameters explicitly.

amannocci:
> Since all the services implemented in this action are implemented already in the octokit, I do not see the need of reimplement them from scratch, we can use the two methods provided by octokit to make all the work.

I agree that whenever possible we should use already existing methods.
However, in that case, we need to craft a whole request body with a lot of parameters.
In my opinion, the GraphQL query is easier to understand and craft than the rest.
In addition, the GraphQL API allows us to only set the lock property instead of pushing all branch protection settings with the REST API.
Both solutions are fine to me since they are safe, documented, and supported by GitHub.

.github/actions/branch-lock/src/main.ts
kuisathaverat: that part I lest to you I give you the API call

amannocci: Sure, I didn't move further yet.
See https://github.com/elastic/apm-pipeline-library/pull/2432#issuecomment-1803732747.
I prefer to wait until we agree on the solution before further development.

amannocci:
We will not use this feature.

PR #2376, opened 2023-09-25
Add pipelineCommit and pipelineBranch inputs to buildkite action

reakaleek:
## What does this PR do?

- Add `pipelineCommit` and `pipelineBranch` inputs to buildkite action.
- deprecates `pipeline

This is in line with the [Buildkite Builds API](https://buildkite.com/docs/apis/rest-api/builds#create-a-build).

## Why is it important?

For testing purposes I was not able to trigger the pipeline from a specific branch, it would always checkout the main.

This gives us the ability to do so.

## Related issues
Closes #ISSUE

PR #2450, opened 2023-11-22
action(buildkite): backward compatible

v1v:
## What does this PR do?

Support the previous behaviour otherwise consumers need to change how they use this github action

## Why is it important?

https://github.com/elastic/apm-pipeline-library/pull/2376 changed the behaviour and `pipelineVersion` was never honoured when it was set by the consumer.

for instance:

```yaml
      - id: buildkite
        continue-on-error: true
        name: Run Deploy
        uses: elastic/apm-pipeline-library/.github/actions/buildkite@current
        with:
          ...
          pipelineVersion: ${{ env.TAG_NAME }}
```

didn't use the correct version defined in `pipelineVersion: ${{ env.TAG_NAME }}` but the default version defined by `pipelineCommit: HEAD` even if `pipelineCommit` was not explicitly used but it was defaulted.


amannocci (approved):
LGTM

PR #2458, opened 2023-12-06
pytest-otel: Fix handling of errors during teardown

schmollf:
## What does this PR do?

Before this patch the pytest-otel plugin would encounter INTERNALERROR if there was a bug during test teardown instead of outputting the proper error. It seems that the 'span[test_name]' expression cannot actually be valid looking at the rest of __init__.py

I also added a test that previously failed and now passes, though I didn't understand why all the tests seem to be duplicated in separate files in the tests/it/ directory. The make target 'make it-test' failed for me due to not having docker-compose.

Testing: 'make test'

## Why is it important?

Without the change test failures during teardown are not properly reported.


schmollf:
Hm I signed the agreement, but the CLA button hasn't picked that up so far. Maybe I need to push a new diff for it to pick that up?

PR #2460, opened 2023-12-09
Pytest otel fix teardown bug

schmollf:
## What does this PR do?

Before this patch the pytest-otel plugin would encounter INTERNALERROR if there was a bug during test teardown instead of outputting the proper error. It seems that the 'span[test_name]' expression cannot actually be valid looking at the rest of init.py

I also added a test that previously failed and now passes, though I didn't understand why all the tests seem to be duplicated in separate files in the tests/it/ directory. The make target 'make it-test' failed for me due to not having docker-compose.

Testing: 'make test'

## Why is it important?

Without the change test failures during teardown are not properly reported.

## Related issues
Closes #ISSUE

### DIFF
--- a/resources/scripts/pytest_otel/src/pytest_otel/__init__.py
+++ b/resources/scripts/pytest_otel/src/pytest_otel/__init__.py
@@ -354,7 +354,7 @@ def pytest_runtest_logreport(report):
     test_name = report.nodeid.split("::")[0]
 
     if report.failed and report.when == "teardown":
-        span = spans[test_name]
+        span = spans[session_name]
         span.set_attribute("tests.systemerr", report.capstderr)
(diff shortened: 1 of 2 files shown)


v1v:
> ❌ Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement):

@schmollf thanks for your contribution, our CLA Checker relies on the git user and so far it seems https://github.com/elastic/apm-pipeline-library/pull/2460/commits/2f96126adb0e4026de9fcf36461250768c9a52f2 has been signed with the email ending with purestorage(dot)com.

Would you mind amending your commits with the same email in https://github.com/elastic/apm-pipeline-library/pull/2460/commits/5e0ee3078bdcc76c9554a00b0c8a207a5a5fcc2a ?

PR #2465, opened 2023-12-12
artifacts-api: corner case when main has been built and previous version no

v1v:
## What does this PR do?

Avoid pointing the previous version to main when main has been moved to another minor version

## Why is it important?

Avoid having wrong versioning

> it’s because the master snapshot was running with 8.12 as a version, we were late to bump it. As soon as a 8.12 snapshot runs it will override the branch


amannocci (approved):
LGTM

PR #2480, opened 2023-12-20
release: set GH_TOKEN, add debug traces and notify if failures

v1v:
## What does this PR do?

Set GH_TOKEN
Add debug traces
Notify if any failures

## Why is it important?

Otherwise, it fails when running on the CI with the error below:

> To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable

### DIFF
--- a/.ci/generate-releases.sh
+++ b/.ci/generate-releases.sh
@@ -215,3 +215,9 @@ if [ "$(isAvailable "$edge_8")" = "true" ] ; then
   # NOTE: when 9.x happens then `main` will point to 9.
   echo "$edge_8" > "$EDGE_FOLDER/main"
 fi
+
+## 3. debug
+if command -v tree ; then
+  echo "The below files have been generated"
+  tree .
+fi
--- a/.ci/generate-releases.sh
+++ b/.ci/generate-releases.sh
@@ -22,6 +22,9 @@
 #
 set -eo pipefail
 
+if [ -n "$RUNNER_DEBUG" ] ; then
+  set -x
+fi
 
 ###############
 ### FUNCTIONS
--- a/.ci/generate-releases.sh
+++ b/.ci/generate-releases.sh
@@ -66,17 +66,6 @@ function latest() {
   rm $file &> /dev/null || true
 }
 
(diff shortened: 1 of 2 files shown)


.github/workflows/generate-elastic-stack-releases.yml
```diff
@@ -45,7 +47,6 @@ jobs:
         uses: 'google-github-actions/upload-cloud-storage@v2'
         with:
           path: releases
-          glob: "**/*.json"
```
v1v, line 48: there are no JSON files...

```bash
.
├── releases
│   ├── current
│   │   ├── 6
│   │   ├── 6.8
│   │   ├── 7
│   │   ├── 7.17
│   │   ├── 8
│   │   └── 8.11
│   ├── edge
│   │   ├── 8
│   │   └── main
│   └── next
│       ├── minor-7
│       ├── minor-7.17
│       ├── minor-8
│       ├── minor-8.12
│       ├── patch-7
│       ├── patch-7.17
│       ├── patch-8
│       └── patch-8.11
└── releases.properties
```

.ci/generate-releases.sh
```diff
@@ -22,6 +22,9 @@
 #
 set -eo pipefail
 
+if [ -n "$RUNNER_DEBUG" ] ; then
```
v1v, line 25: help debugging when running in a debug mode through the [GitHub actions](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging)

.ci/generate-releases.sh
```diff
@@ -172,7 +175,7 @@ cd releases
   echo "next_patch_8=$next_patch_8"
   echo "edge_8=$edge_8"
   echo "generated=https://github.com/elastic/apm-pipeline-library/actions/workflows/generate-elastic-stack-releases.yml"
-} > releases.properties
+} | tee releases.properties
```
v1v, line 178: print something in the console while this runs so we can see what versions will be shown.

amannocci (approved):

PR #2487, opened 2024-01-08
Revert "chore(deps): bump the github-actions group in /.github/actions/buildkite with 1 update"

v1v:
Reverts elastic/apm-pipeline-library#2474

Otherwise, consumers need to use v4 and we have not done that yet


reakaleek (approved):
Lgtm

I don't understand. How are consumers affected by this?

Or is it failing?

Or rather, what is the reason that consumers must use v4 in another step/job?

PR #2544, opened 2024-02-19
Added action to wait for maven-central artifacts to be available

JonasKunz:
## What does this PR do?

Adds an action which allows to wait for a certain maven-central artifact to be available.

I've tested the action in a dummy-repo on my account.

## Why is it important?

When releasing artifacts to be maven central, the output of the release task is usually not useful: sometimes the output says failed, even though the artifact was published. Sometimes the release task is successfull, but the artifact wasn't published.

For this reason it is beneficial to just ignore the output of the release step and instead check the presence of the desired artifact on maven central, which is accomplished using this task.

In addition in the future we could enhance this task to provide the artifact as an output, so that we can for example publish it as artifact on the GH release.

### DIFF
--- a/.github/actions/await-maven-artifact/action.yml
+++ b/.github/actions/await-maven-artifact/action.yml
@@ -1,0 +1,24 @@
+name: 'Wait for Maven Central Artifact'
+description: 'Waits for an artifact to be available on maven central'
+inputs:
+  groupid:  
+    description: 'Maven group-ID of the artifact'
+    required: true
+  artifactid:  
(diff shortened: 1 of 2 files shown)


v1v (approved):
Awesome! Thanks

v1v (approved):

v1v:
I've just released a new version for this library and you can use `@current`:

```yaml
jobs:
  wait-release:
    runs-on: ubuntu-latest
    steps:
      - uses: elastic/apm-pipeline-library/.github/actions/await-maven-artifact@current
        with:
          ...
```
```

PR #2553, opened 2024-03-06
oblt-cli: pin version 6.5.3

v1v:
## What does this PR do?

PIn the version to `6.5.3`

## Why is it important?

Prepare for the secrets system migration so consumers don't get affected by the change.

## Follow-ups

Migrate to the new secret system store


v1v:
linting is a false positive and unrelated!

v1v:
@mergifyio rebase

PR #2582, opened 2024-04-15
github-action(kibana-docker-image): skip CDN

v1v:
## What does this PR do?

Disable the CDN thing for Kibana

## Why is it important?

Otherwise, serverless deployments use the CDN thing... 

## Related issues

Caused by https://github.com/elastic/kibana/pull/169707


reakaleek (approved):

kuisathaverat (approved):
the image will fail to show some assets in any case

v1v:
Let's merge it anyway, the issue is somewhere else but we don't need those assets and it might run faster

PR #2583, opened 2024-04-16
github-action: add label when deploy-my-kibana

v1v:
## What does this PR do?

Initial attempt to integrate the Deploy My Kibana with the existing Kibana CI mechanism

## Why is it important?

Reuse what the Kibana CI generates


reakaleek (approved):

v1v:
we use a different approach

PR #2596, opened 2024-05-03
feat(cluster-name): support oblt-cli with cluster-name flag

v1v:
## What does this PR do?

Add support for the `--cluster-name` when using the `oblt-cli`

## Why is it important?

Being able to fix the name of the ephemeral cluster, so the updates can work as expected.

### DIFF
--- a/.github/actions/oblt-cli-create-custom/action.yml
+++ b/.github/actions/oblt-cli-create-custom/action.yml
@@ -10,6 +10,9 @@ inputs:
   token:
     description: 'The GitHub access token.'
     required: true
+  cluster-name:
+    description: 'Name of the cluster if no a random one.'
+    required: false
   cluster-name-prefix:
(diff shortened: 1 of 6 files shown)


.github/actions/oblt-cli-create-custom/action.yml
```diff
@@ -72,6 +75,8 @@ runs:
       run: |
         {
           echo -n '--template=${{ inputs.template }} '
+          [ -n '${{ inputs.cluster-name }}' ] \
```
kuisathaverat, line 78: Possible injection

v1v: I'll be raising a follow-up to avoid the injection, rather than changing here the `cluster-name` 👼

reakaleek (approved):

PR #2603, opened 2024-05-09
jenkins: disable send telemetry data by default

v1v:
## What does this PR do?

Disable sending telemetry dta

## Why is it important?
Jenkins sunset, so we can decommission the infrastructure

## Related issues
Closes #ISSUE

### DIFF
--- a/.github/workflows/pytest_otel.yml
+++ b/.github/workflows/pytest_otel.yml
@@ -36,7 +36,7 @@ jobs:
         run: make -C resources/scripts/pytest_otel test
       - name: it-test
         run: make -C resources/scripts/pytest_otel it-test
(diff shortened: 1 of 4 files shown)


.github/workflows/pytest_otel.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         run: make -C resources/scripts/pytest_otel test
       - name: it-test
         run: make -C resources/scripts/pytest_otel it-test
-      - uses: actions/upload-artifact@v2  # upload test results
+      - uses: actions/upload-artifact@v4  # upload test results
```
v1v, line 39: unrelated but somehow the linting ran so I decided to fix this

amannocci (approved):

PR #2608, opened 2024-05-13
Test oblt-cli google auth

reakaleek:
## Related issues
Closes #ISSUE

### DIFF
--- a/.github/actions/setup-oblt-cli/action.yml
+++ b/.github/actions/setup-oblt-cli/action.yml
@@ -45,7 +45,7 @@ runs:
     - name: Create workloadIdentityPool provider ID
       id: generate-workload-identity-pool-provider-id
       run: |
-        HASH=$(echo "${{ github.repository }}" | sha256sum | cut -c1-22)
+        HASH=$(echo -n "${{ github.repository }}" | sha256sum | cut -c1-22)
         echo "workload_identity_provider_id=oblt-cli-${HASH}" >> "${GITHUB_OUTPUT}"
       shell: bash
     - uses: 'google-github-actions/auth@v2'
--- a/.github/actions/setup-oblt-cli/action.yml
+++ b/.github/actions/setup-oblt-cli/action.yml
@@ -46,6 +46,7 @@ runs:
       id: generate-workload-identity-pool-provider-id
       run: |
         WORKFLOW_PATH=$(echo "${GITHUB_WORKFLOW_REF}" | awk -F@ '{print $1}')
+        echo "##[debug] ${WORKFLOW_PATH}"
         HASH=$(echo -n "${GITHUB_REPOSITORY}/${WORKFLOW_PATH}" | sha256sum | cut -c1-23)
         echo "workload_identity_provider_id=workflow-${HASH}" >> "${GITHUB_OUTPUT}"
       shell: bash
--- a/.github/actions/setup-oblt-cli/action.yml
+++ b/.github/actions/setup-oblt-cli/action.yml
@@ -16,7 +16,7 @@ inputs:
     required: false
   version:
     description: "Install a specific version of oblt-cli"
-    default: "6.5.3"
+    default: "7.1.3"
     required: false
 runs:
   using: composite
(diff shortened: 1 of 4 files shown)


.github/actions/setup-oblt-cli/action.yml
```diff
@@ -42,3 +42,13 @@ runs:
...
+    - name: Create workloadIdentityPool provider ID
+      id: generate-workload-identity-pool-provider-id
+      run: |
+        HASH=$(echo "${GITHUB_WORKFLOW_REF}" | awk -F@ '{printf $1}' | sha256sum | cut -c1-23)
+        echo "workload_identity_provider_id=workflow-${HASH}" >> "${GITHUB_OUTPUT}"
+      shell: bash
+    - uses: 'google-github-actions/auth@v2'
+      with:
+        project_id: 'elastic-observability'
+        workload_identity_provider: 'projects/8560181848/locations/global/workloadIdentityPools/github/providers/${{ steps.generate-workload-identity-pool-provider-id.outputs.workload_identity_provider_id }}'
```
v1v, line 54: Do you think we can provide a GitHub composite action in charge of doing this? So any consumers can benefit from?

Or maybe that's something to be implicitly called by the `setup-oblt-cli`. What do you think?

v1v: I know this is for testing purposes :)

reakaleek: I was actually thinking of creating a wrapper google auth action with this code.

which can be used at the beginning of every job, where it's needed.

reakaleek: Another Idea I had.. was to save the workload identity id as a GitHub secret.

This way.. only workflows with access to secret could authenticate.

Of course, this assumes that the ID is obscured and not backtrackable—security by obscurity.

The algorithm to create the ID would need to change. Since this PR is already public 😅

v1v: We can start with some simple.

I think the wrapper is neat, and maybe `setup-oblt-cli` can use it if no vault inputs are passed?

reakaleek: I don't think we should use it by default.

I think it's a fair use case to have other resources than just oblt-cli and authenticating.

So, IMO, authenticating should happen outside of the setup-oblt-cli step.

However, maybe we can have a convenience flag to do the authentication?

v1v: I think `oblt-cli` is quite an opinionated tool that requires Google authentication or vault access, regardless. IMO, consumers should not care about this, for the shake of simplicity.

I'd expect a simple way of using it like:

```yaml
jobs:
  run-oblt-cli-create-serverless:
    runs-on: ubuntu-latest
    steps:
      - uses: ./.github/actions/oblt-cli-create-serverless
        with:
          token: ${{ secrets.MY_TOKEN }}
          target: 'qa'
          cluster-name-prefix: 'testgithubaction'
```

rather than

```yaml
jobs:
  run-oblt-cli-create-serverless:
    runs-on: ubuntu-latest
    steps:
      - uses: ./.github/actions/setup-auth
         with:
            ....
      - uses: ./.github/actions/oblt-cli-create-serverless
        with:
          token: ${{ secrets.MY_TOKEN }}
          target: 'qa'
          cluster-name-prefix: 'testgithubaction'
```

> However, maybe we can have a convenience flag to do the authentication?

I don't envision we will use it with default `false`, therefore we will need to set it to true explicitly.

Do you envision a different use case?

reakaleek: So, my thinking is.

The more we use keyless, the more resources we will need.

Hence, there will be workflows where we use both oblt-cli, or terraform, or gcloud.

Then the question arises, when should we do the auth step?

I think it can lead to confusion if oblt-cli is authenticating by default.

It could break existing authentication attempts in previous steps.

I think it could be a similar  reason, why these actions are separated e.g. `google-github-actions/setup-gcloud` and ``google-github-actions/auth`

reakaleek: Separating this intentionally makes it implicit, and having docs that show the requirement for this.. could be enough.

reakaleek: Generally, I agree with you.. making authenticate by default is the better user experience.


But I think a strong requirement for this is.. that we don't mess up existing authentications to GCP.

reakaleek: I will dig a bit more into this.

reakaleek: Unfortunately, I was not able to find a way to authenticate in an isolated way.

PR #2639, opened 2024-06-25
Add deprecation warning

v1v:
## What does this PR do?

Create warning with the alternative

## Why is it important?

Help consumers our strategy

## Test

See 
image

### DIFF
--- a/.github/actions/updatecli/action.yml
+++ b/.github/actions/updatecli/action.yml
@@ -46,7 +46,7 @@ runs:
     - uses: actions/github-script@v7
       with:
         script: |
-          core.warning('This action is deprecated. Use `updatecli/updatecli-action` with GitHub secrets.')
+          core.warning('This action is deprecated. Use https://github.com/elastic/oblt-actions/tree/main/updatecli/run.')
 
     - name: Setup Git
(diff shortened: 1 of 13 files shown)


.github/actions/updatecli/action.yml
```diff
@@ -43,6 +43,11 @@ runs:
   using: "composite"
   steps:
 
+    - uses: actions/github-script@v7
+      with:
+        script: |
+          core.warning('This action is deprecated. Use `updatecli/updatecli-action` with GitHub secrets.')
```
reakaleek, line 49: What about https://github.com/elastic/oblt-actions/tree/main/updatecli/run ?

reakaleek (dismissed):

.github/actions/updatecli/action.yml
v1v: good catch! I missed it

v1v:
```suggestion
          core.warning('This action is deprecated. Use https://github.com/elastic/oblt-actions/tree/main/updatecli/run.')
```

reakaleek (approved):

